PR #121, opened 2022-12-29
Make each policy provision a sentence #88

### LINKED ISSUE #88: Make each policy provision a sentence
How about replacing the current with this in the right panel (soon to be carousel thanks to @DeepthiSRao)?
>Raise the Ofgem Energy Price Cap from **£70** to **£80**
_2022-01-01 to 2027-12-31_

Where £70 could be gray and £80 could be blue

DeepthiSRao:
Fixes #88

### DIFF
--- a/src/pages/policy/output/PolicyOutput.jsx
+++ b/src/pages/policy/output/PolicyOutput.jsx
@@ -83,6 +83,15 @@ export default function PolicyOutput(props) {
       asyncApiCall(url, null, 3_000)
         .then((data) => {
           if (data.status === "error") {
+            if (!data.result.baseline_economy) {
+              data.result.baseline_economy = { status: "error", message: "An error outside the baseline economy computation." };
+            }
+            if (!data.result.reform_economy) {
+              data.result.reform_economy = { status: "error", message: "An error outside the baseline economy computation." };
+            }
+            if(data.message) {
+              data.result.message = data.message;
+            }
             setError(data.result);
           } else {
             setImpact(data.result);
@@ -152,6 +161,16 @@ export default function PolicyOutput(props) {
                 <CloseCircleFilled style={{fontSize: 20, color: "red"}} />
             }
           />
+          {
+            error.message ?
+              <Steps.Item
+                title="Comparison"
+                description={error.message}
+                status="wait"
+                icon={<CloseCircleFilled style={{fontSize: 20, color: "red"}} />}
+              /> :
+              null
+          }
         </Steps>
         </div>
       }


src/pages/policy/PolicyRightSidebar.jsx
```diff
@@ -41,29 +40,26 @@ function PolicyNamer(props) {
...
+      style={{ 
+        display: "flex", 
+        flexDirection: "column",
+        alignItems: "flex-start",
+        padding: "0 10%" }}
+    >
+      <div style={{ textAlign: "left" }}>
+          {paramLabel} from{' '}
+          <span className="text-secondary fw-bold">{oldValStr}</span> to{' '} 
+          <span className="text-primary fw-bold">{newValueStr}</span>
```
nikhilwoodruff, line 59: Could we avoid colouring the text blue?

src/pages/policy/PolicyRightSidebar.jsx
```diff
@@ -41,29 +40,26 @@ function PolicyNamer(props) {
...
-        <div>until {endDate}</div>
+    <div 
+      style={{ 
+        display: "flex", 
+        flexDirection: "column",
+        alignItems: "flex-start",
+        padding: "0 10%" }}
+    >
+      <div style={{ textAlign: "left" }}>
+          {paramLabel} from{' '}
```
nikhilwoodruff, line 57: Could we add a `raise`/`lower`/`change` prefix here?

nikhilwoodruff (changes requested):
Thanks for this @DeepthiSRao ! Added a few comments, and here's what it looks like on my machine for the thread: 
image

src/pages/policy/PolicyRightSidebar.jsx
nikhilwoodruff: Ah sorry ignore this - just remembered we did decide to make this blue. Though could we use the same blue as the rest of the site? It's in `style.colors.BLUE`.

src/pages/policy/PolicyRightSidebar.jsx
DeepthiSRao: @nikhilwoodruff Thank you. 
Are you asking me to name any styling prefix or for the value to be displayed? Could you please give an example.

nikhilwoodruff: Ah, I meant adding a prefix to the text so that the message becomes:
`Raise [basic rate] from 20% to 40%`
or
`Lower [higher rate] from 40% to 35%`
or
`Change [refundable CTC] from false to true`

Depending on whether the new value is (a number and higher), (a number and lower), or (not a number/a boolean). Does that make sense?

MaxGhenis: How about `[Disable/Enable] refundable CTC` instead of `Change refundable CTC from false to true`?

PR #128, opened 2022-12-30
Make each policy provision a sentence #88

### LINKED ISSUE #88: Make each policy provision a sentence

DeepthiSRao:
Fixes #88


MaxGhenis:
Could you display `Enable` if `value` is `true`, otherwise `Disable`?

DeepthiSRao:
PE_Issue88_1
PE_Issue88_2

MaxGhenis:
Please remove the "from..." on the enable/disable sentences

It should just be "Enable abolition" or "Disable abolition" with Enable/Disable in bold+blue

(Separately, we should make this particular parameter more descriptive...)

MaxGhenis:
Thanks - last thing, could you please center it? I think that's usually the more natural layout when there's a carousel.

DeepthiSRao:
image

PR #179, opened 2023-01-21
Use bold instead of arrows to indicate active element of left menus

### LINKED ISSUE #170: Use bold instead of arrows to indicate active element of left menus
For example, `Excess shelter expense` is the same level as `Excess medical expense`, and it is the parent of `SNAP maximum shelter deduction`. The arrows do not clearly indicate this. Bold is the UI standard.
image

Felix-Cadiz:
Fixes #170

1) Parent tabs are now consistently in line with one another.
2) Arrows are removed.
3) Current tabs are now bolded.
4) Hovered tabs are given a light bolding; however, they are not as dark as the current tab.

Screen Shot 2023-01-20 at 4 13 43 PM

### DIFF
--- a/src/layout/Menu.jsx
+++ b/src/layout/Menu.jsx
@@ -20,10 +20,10 @@ function MenuItem(props) {
       <motion.h5
         style={{ fontSize: 18 }}
         initial={{ x: 0 }}
-        whileHover={{ x: 5 }}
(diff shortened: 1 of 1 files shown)


nikhilwoodruff:
Thanks @Felix-Cadiz - this looks great now!

PR #187, opened 2023-01-26
Optimise images

nikhilwoodruff:


MaxGhenis (commented):
How did you do this optimization?

PR #220, opened 2023-02-09
Expand x axis to max(200k, earnings*2)

MaxGhenis:
Other changes showed up via my formatter

Example with $150k income:
image
image


nikhilwoodruff (approved):
Thanks- LGTM!

PR #221, opened 2023-02-09
Added subscribe button

nehashinde9918:


nikhilwoodruff:
image
Looks good to me, thanks @nehashinde9918! I think we can merge this but filed #223 since it would be a nice-to-have but seems not immediately obvious how to do it.

PR #253, opened 2023-02-15
Issue #249: Extend 404 to policyengine.org/{us,uk}/[bad-url]

### LINKED ISSUE #249: Extend 404 to policyengine.org/{us,uk}/[bad-url]
PolicyEngine/policyengine#932 adds 404 redirects from policyengine.org/[bad-url], let's also add the country-specific pages

PGrad:
# New feature/Improvement/Bug fix: Adds 404 redirect for country pages
 
When the user enters a bad url for a country page like `policyengine.org/uk/nonsense`, they will now be redirected to a 404 page. If the user tries to navigate to a non-existent country page, like `policyengine.org/flatland`, they will be redirected to the UK page by default. I used [this PR](https://github.com/PolicyEngine/policyengine/pull/932) from the PolicyEngine repo as a reference. 

Prior to this submission, if you entered a bad country like `policyengine.org/flatland` you would see a blank page with no feedback as to what the user did wrong. I redirect to the UK page because it's simpler then showing the FOF 404 page, given that I would need to either add a header that is conditionally displayed (so `policyengine.org/uk/nonsense` doesn't show 2 headers), or move the header in the country pages up a level.

### Status
* [x]  Cosmetic change
   
* [x]  Screenshots attached
PolicyEngine
The url I used here was "policyengine.org/uk/nonsense"

* [ ]  Back-end change
* [ ]  Unintuitive logic commented
* [ ]  Version change   
  * [ ]  Major
  * [ ]  Minor
  * [ ]  Patch
 
### Issues fixed
Closes #249

### DIFF
--- a/src/PolicyEngine.jsx
+++ b/src/PolicyEngine.jsx
@@ -21,7 +21,7 @@ export default function PolicyEngine() {
         <Route path="/us/*" element={<PolicyEngineCountry countryId="us" />} />
         <Route path="/ca/*" element={<PolicyEngineCountry countryId="ca" />} />
         <Route path="/ng/*" element={<PolicyEngineCountry countryId="ng" />} />
-        <Route exact path="/" element={<Navigate to={`/${countryId}`} />} />
+        <Route path="/*" element={<Navigate to={`/${countryId}`} />} />
       </Routes>
     </Router>
   );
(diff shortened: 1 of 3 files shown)


nikhilwoodruff (approved):
Thanks @PGrad! This looks good to merge, your finding with the complexity around {bad-country} makes sense. I think we probably do want to show a 404 page for those URLs though (given the /uk endpoint is most likely unexpected in that scenario), so I filed #254 but I think we can merge this in the meantime. Thanks for your contribution!

PR #255, opened 2023-02-16
Add tests for issue #249

PGrad:
# New feature/Improvement/Bug fix: Adds tests for issue #249 

This submission adds some tests in jest that confirm the current behavior implemented in #249: that the user is directed to the UK page if they enter an invalid country code, and directed to a 404 page if they give a bad url under one of the country pages. Tests ensure that the website continues to behave as expected.

Adds jest, react-testing-library, and jsdom as devdependencies in the package.json. This submission also adds a new Github action that automatically runs the tests when pushing to a PR.

### Status
* [x]  Cosmetic change
Moves the top margin of the 404 page up a bit.
   
* [ ]  Screenshots attached
* [ ]  Back-end change
* [ ]  Unintuitive logic commented
* [ ]  Version change   
  * [ ]  Major
  * [ ]  Minor
  * [ ]  Patch
 
### Issues fixed
None

Issue #210, opened 2023-02-06
Logo off center in blog on mobile 


MaxGhenis:
![Screenshot_20230206-085938.png](https://user-images.githubusercontent.com/6076111/216991058-57581ceb-ecb1-43b0-b24e-e113b46c855d.png)

PGrad:
This looks like a hard one but I'm on it. It doesn't reproduce if I use the mobile emulator in the inspector on my laptop.

PR #266, opened 2023-02-25
Issue #210: Logo off center in blog on mobile

PGrad:
#### Fixing issue #210
```
src/pages/BlogPage.jsx
```
On first examination, it appears that the logo is off center on mobile. However, the header is just fitting itself to the dimensions of the BlogPage component beneath it. On mobile, the iframe element that contains video was overflowing the boundaries of the viewport, which meant the BlogPage component as a whole was overflowing and the header expanded itself in tandem. My solution is to force the iframe to fit the viewport width on mobile, which may cause it to break its original aspect ratio but you can still see the preview and play the video just fine:

VideoPreview

Now that the BlogPage fits the viewport, the header is centered.

#### Fixing alignment issue for multiple authors
```
src/pages/BlogPage.jsx
```
I also fixed an issue where if you had multiple authors on a blog, their respective `AuthorSection` elements would display side-by-side and also overflow the page. An easy fix is to render them in a column, one below the other. The `AuthorSection` elements were also not aligned using Flexbox, so I use CSS Grid instead to get exact control over how much space every sub-element takes.
Authors

#### Limit CI/CD to node version 19
```
.github/workflows/pr.yml
package.json
package-lock.json
```
Nikhil suggested in issue #255 that we only test for node version 19, and so I changed the Github action to only test on that version. I also require node version 19 in the package.json, so if a developer is not on 19 they will be encouraged to move onto it.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,7 +24,7 @@ jobs:
         # by our package.json.
         # Hopefully this allows us to catch problems
         # with differing versions ahead of time.
-        node-version: [16.x, 17.x, 18.x, 19.x]
+        node-version: [19.x]
       
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


nikhilwoodruff (approved):
All looks good to me, merging. Thanks @PGrad!

PR #278, opened 2023-02-28
Utilize Hovercard component in Household Charts

### LINKED ISSUE #262: Add nice hover cards from policy reform charts to household charts

samau3:
Fixes #262 

image

### DIFF
--- a/src/pages/household/output/EarningsVariation/BaselineAndReformChart.jsx
+++ b/src/pages/household/output/EarningsVariation/BaselineAndReformChart.jsx
@@ -199,7 +199,9 @@ function BaselineAndReformTogetherChart(props) {
             currency:"USD",
             maximumFractionDigits: 0
           });
-        const message = `Net Income ${netIncome} Employment Income ${employmentIncome}`
+        const message = netIncome && employmentIncome 
+          ? `Net Income ${netIncome} Employment Income ${employmentIncome}`
+          : ""
         setHoverCard({
           title: data.points[0].x === +currentEarnings 
             ? `Your current ${variableLabel}` 
@@ -235,6 +237,7 @@ function BaselineReformDeltaChart(props) {
       line: {
         color: style.colors.BLUE,
       },
+      hoverinfo:"none",
     },
     {
       x: [currentEarnings, currentEarnings],
@@ -244,6 +247,7 @@ function BaselineReformDeltaChart(props) {
       line: {
         color: style.colors.MEDIUM_DARK_GRAY,
       },
+      hoverinfo:"none",
     },
   ];
   const plotObject = (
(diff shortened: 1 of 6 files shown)


src/pages/household/output/EarningsVariation/BaselineAndReformChart.jsx
```diff
@@ -184,10 +188,29 @@ function BaselineAndReformTogetherChart(props) {
...
+          { style:"currency",
+            currency:"USD",
+            maximumFractionDigits: 0
+          });
+        const employmentIncome = data.points[0].x.toLocaleString("en-US", 
+          { style:"currency",
+            currency:"USD",
+            maximumFractionDigits: 0
+          });
+        const message = `Net Income ${netIncome} Employment Income ${employmentIncome}`
```
samau3, line 202: I tried adding a new line character `\n` to make Net Income and Employment Income appear on different lines on the Hovercard, but that had no effect on the displayed text. I also double checked the string that's passed to the Hovercard component if a `\n` is included  and confirmed with a `console.log` that the new line character is passed to the component correctly, despite the label text not displaying it.

nikhilwoodruff: Yeah I think we might need to actually use a <br> tag for this rather than just a newline character?

src/pages/household/output/EarningsVariation/BaselineOnlyChart.jsx
```diff
@@ -110,8 +115,29 @@ export default function BaselineOnlyChart(props) {
         style={{
           width: "100%",
         }}
+        onHover={(data) => {
+          const netIncome = data.points[0].y?.toLocaleString("en-US", 
+            { style:"currency",
+              currency:"USD",
```
nikhilwoodruff, line 121: Does this show pounds for the UK view?

nikhilwoodruff:
Thanks @samau3, this looks really neat? Just left a few comments.

src/pages/household/output/EarningsVariation/BaselineAndReformChart.jsx
samau3: Thanks for the suggestion! Utilizing a `<br>` tag worked in splitting the string into the two lines.

src/pages/household/output/EarningsVariation/BaselineOnlyChart.jsx
samau3: Since it was hard coded as `USD` this did not properly show pounds for the UK view. Following how the policy charts handled displaying currency, I made the change of appending the `metadata.currency` character to the string returned from the `.toLocaleString` method.

MaxGhenis:
Let's change the wording to:
>If you earn $x, your {variable} will be $y

nikhilwoodruff:
This looks amazing so far @samau3! Just a few nitpicks (last ones I think) I found when playing around with it:

* On the baseline-reform chart, the baseline line still shows the default Plotly hovercard as well as the new one
* On that chart also, the text uses the variable name rather than the label (e.g. your **household_net_income** is ... rather than your **net income** is)

samau3:
> * On the baseline-reform chart, the baseline line still shows the default Plotly hovercard as well as the new one

 I added an extra `hoverinfo: none` on the Baseline chart that I missed.

> * On that chart also, the text uses the variable name rather than the label (e.g. your **household_net_income** is ... rather than your **net income** is)

Looks like I was using the wrong variable, where I used `variable` instead of `variableLabel` - made the change!

MaxGhenis:
Sometimes taxes contribute to cliffs as well as public benefits, so how about this?
>Your net income falls after earning $minIncome until earning $maxIncome.

And if there's a reform, it could be
>Your net income falls after earning $minIncome until earning $maxIncome in the {baseline,reform} scenario.

nikhilwoodruff (approved):
This looks great when testing it out! Just added a very minor fix I spotted (the cliff hovercard read 'in the null scenario' for me, but just needed a small tweak). Looks good to merge, thanks @samau3!

PR #291, opened 2023-03-03
Add linting and prettier

### LINKED ISSUE #290: Add linting and prettier
This issue will add git hooks for linting and formatting the code.

PGrad:
Fixes #290 

### Summary
To improve the quality of the code, I added [`eslint`](https://www.npmjs.com/package/eslint) and [`prettier`](https://prettier.io/) to our dev dependencies, and then added git hooks which will lint and format the code. I use [`husky`](https://typicode.github.io/husky/#/) to create the git hooks.

The approach here is to lint and format the code gradually, as every commit we make is going to clean up the code file by file on pre-commit, and then eventually require the linter pass on pre-push and on CI. A further step would be to add airbnb style to the linter so we can have a single enforced style.

### New git hooks
```
Linter config
.eslintrc.json
jest.config.js

Husky pre and post commits
.husky/post-commit
.husky/pre-commit

Prettier config
.prettierignore
.prettierrc.json

Dependency updates
package-lock.json
package.json
```
There are 2 new git hooks added: a pre-commit and post-commit hook. The pre-commit hook will format any staged files with prettier using [`lint-staged`](https://github.com/okonet/lint-staged) _before_ the commit. For example, if you make changes to package.json and stage it (i.e. `git add`), when you commit package.json will be formatted by prettier and that new formatted version of the file will be what's committed. This means that you can write your code without having to format things manually, but it's still preferable to use an editor extension (for example with [VS Code](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)).

eslint will also be run on your staged files on pre-commit and the commit will fail if there's an error.

The post-commit hook will check if you have any untracked files. I have a habit of making a commit without adding new files, so it'd be nice to avoid doing that when contributing here.

### Linter fixes
```
src/
    - PolicyEngineCountry.jsx

    api/
        - call.js
        - variables.js
    pages/household/input/
        - CountChildren.jsx
        - MaritalStatus.jsx
        - VariableEditor.jsx
```
Some files were updated to fix linter errors that popped up when running `npm run lint`

### Note to Developers
On a first run of `npm start`, you may see a lot of warnings pop up, but they're not going to stop the build. You can ignore them and eventually we'll try to whittle them down.

If you want to run the linter locally, do `npm run lint`.


nikhilwoodruff (approved):
This all looks great, thanks @PGrad. Verified it works on my machine.

Issue #297, opened 2023-03-07
New household chart hovercards prevent zooming


MaxGhenis:
[Example](https://policyengine.org/us/household?focus=householdOutput.earnings&household=11373)

https://www.loom.com/share/d72a5e5cb5b54139b64650eb41a51056

MaxGhenis:
Fixed in #298

PR #298, opened 2023-03-07
Resolve zooming issue in household charts

samau3:
Fixes issue #297.

Previously, zooming into a chart and hovering over the line caused the entire Plot component to re-render, resetting the  view back to the default zoom. By utilizing `uirevision` based on this [SO discusion](https://stackoverflow.com/questions/63519850/plotly-react-keep-zoom-and-rotation-between-re-rendering), the charts are able to keep their zoomed in view while hovering over the plot lines. 

Screen Recording 2023-03-07 at 1 40 44 PM


nikhilwoodruff (approved):
Thanks @samau3 for finding the answer to this! Looks great, merging.

PR #300, opened 2023-03-08
Fix: Only allow numeric entry

### LINKED ISSUE #143: Only allow numeric entry
For example a user entered a comma in the employment income box and got this
image

DeepthiSRao:
Fixes #143


nikhilwoodruff:
Thanks @DeepthiSRao! Just testing this out, I noticed that the default behaviour when entering a string seems to be to enter in a value of zero. I think the best behaviour here would be to catch the input before it's entered and cancel it if it doesn't pass the test: I think this should be achievable if we add a check in the `onKeyUp` handler, inside the if condition that filters for the `Enter` key? Other than that looks great, good to merge with this change applied.

DeepthiSRao:
Added number format validation for keyup event.

PR #304, opened 2023-03-09
Logo Centering on Homepage

### LINKED ISSUE #73: Top logo is off center on mobile homepage
Looks like it's centered with respect to the space not taken by the flag drop-down

Felix-Cadiz:
Fixes #73

I am having difficulty finding where the country selector is removed from the front end. As it currently stands, the mobile logo will be centered, but everything within the app will be off center.

### DIFF
--- a/src/layout/Header.jsx
+++ b/src/layout/Header.jsx
@@ -34,7 +34,6 @@ function MobileHeader(props) {
           display: "flex",
           backgroundColor: style.colors.BLUE,
           height: HEADER_HEIGHT,
-          paddingLeft: CountrySelector ? 50 : 10,
           position: "fixed",
           width: "100%",
           zIndex: 100,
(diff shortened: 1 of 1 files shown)


Felix-Cadiz:
Logo in Home.
Screen Shot 2023-03-08 at 4 10 47 PM

Logo elsewhere in the app.
Screen Shot 2023-03-08 at 4 10 53 PM

Felix-Cadiz:
To avoid the logo from shifting when the country select is present, I moved the country select outside of the flexbox and hard coded its position. As a result, adjustments will need to be made if we change the header size or the logo size. 

Screen Shot 2023-03-09 at 9 39 02 AM
Screen Shot 2023-03-09 at 9 39 17 AM

nikhilwoodruff (approved):
Thanks @Felix-Cadiz - this looks good to me, no obvious real issues here with absolute positioning.

PR #307, opened 2023-03-09
Compare use of hovertemplate instead of custom Hovercard

samau3:
Explored the use of plotly's `hovertemplate` property for creating custom hover labels on the Household charts.

Hovertemplate hover labels

One drawback to using `hovertemplate` is that the [styling options](https://plotly.com/javascript/reference/layout/#layout-hoverlabel) are more limited.

### DIFF
--- a/src/pages/household/output/EarningsVariation/BaselineAndReformChart.jsx
+++ b/src/pages/household/output/EarningsVariation/BaselineAndReformChart.jsx
@@ -137,7 +137,10 @@ function BaselineAndReformTogetherChart(props) {
       line: {
         color: style.colors.MEDIUM_DARK_GRAY,
       },
-      hoverinfo:"none",
+      // hoverinfo:"none",
+      hovertemplate: `<b>Baseline ${variableLabel}</b><br><br>` +
+                      `If you earn %{x}, your` +
+                      `<br>${variableLabel} will be %{y}.</br><extra></extra>`
     },
     {
       x: earningsArray,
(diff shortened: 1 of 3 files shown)


MaxGhenis:
Thanks Sammy, I like it, especially how it snaps to the line. The amount of code looks pretty comparable too. @nikhilwoodruff any concerns moving to this?

nikhilwoodruff:
Thanks @samau3 for this- so I guess my only real concern is on the amount of control we have over the actual text. For example, if we wanted to write "At X earnings, your income would rise/fall by Y", I'm not sure we'd be able to in the `hovertemplate` field? Or at least not without adding a `customdata` field with all the corresponding hovertexts. And while I don't think that applies to all hovertexts we might use, there then becomes the issue of whether it's best to have the same hovercard UI across all charts or not (I think it is but not *super* importantly). Open to discussion though

MaxGhenis:
Let's also consider the reproducibility in the (to-be-created) `policyengine` Python package. Does plotly.py support our current custom Hovercard?

nikhilwoodruff:
OK, updating on this:
* Could we explore how feasible it is to have the existing hovercards snap to their targets like Plotly? If it is simple, I'd favour that (filed https://github.com/PolicyEngine/policyengine-app/issues/373)
* Reproducibility in the policyengine package is nice, but IMO a) a far less common use-case and b) I'm sure we could (without much more complexity) get the custom hovercards working in that given it's just an export to HTML/JS tags at the end of the day

MaxGhenis:
Two other issues I think might be related to the custom hovercard:
1. "Shaking" #504 
2. Hovercards don't show up on mobile

MaxGhenis:
Let's revive this given the shaking charts. cc @jacky5124

anth-volk:
Seems to me we've basically taken this and incorporated it into multiple past commits. Closing.

PR #312, opened 2023-03-10
Fix some input numbers being formatted as $012000 instead of $12,000 in net income table

### LINKED ISSUE #129: Some input numbers are formatted as `$012000` instead of `$12,000` in net income table
https://www.loom.com/share/f834301aa51b44579828a549fac57132

image

samau3:
Fixes #129.

Previously, inputs for income were being submitted as a type `string`. By changing the submitted value to a valid type `number`, this enables the Net Income output page to properly display the income amounts as a converted currency string. 

image


nikhilwoodruff (approved):
Aha, thanks @samau3! Merging.

PR #315, opened 2023-03-11
Add Oregon exemption post

kfoster19:


MaxGhenis (commented):
Some weird characters in here like `OregonΓÇÖs`, did you just use the mediumexporter?

@nikhilwoodruff do you know if anyone's used mediumexporter on Windows?

PR #319, opened 2023-03-13
Fix Issue 314: Reproduce in Python doesn't capture US state

### LINKED ISSUE #314: `Reproduce in Python` doesn't capture US state
For example, [this household](https://policyengine.org/us/household?focus=householdOutput.pythonReproducibility&household=13191) is set to Alaska, but the `household` bit has no state specified.

>situation = { "people": { "you": {} }, "families": { "your family": { "members": [ "you" ] } }, "marital_units": { "your marital unit": { "members": [ "you" ] } }, "tax_units": { "your tax unit": { "members": [ "you" ] } }, "spm_units": { "your household": { "members": [ "you" ] } }, "households": { "your household": { "members": [ "you" ] } } }

samau3:
Fixes #314.

The `optimiseHousehold` function removes extraneous object keys that already exist in the metadata, allowing for the `Reproduce in Python` page to only have the unique inputs required to replicate the results. 

However, the `state_name` property in the metadata has a `defaultValue` value of `CA` and an `isInputVariable` value of `false`. This caused the `optimiseHousehold` function to remove `state_name` from the final returned object. To avoid this without any major changes to the function inputs, I made a change similar to how the `members` variable is handled. 

```
if (variable === "members" || variable === "state_name")
```

With this change, the `state_name` property is included as a value in the `households` key.
```
//...
situation = {
  //...
  "households": {
    "your household": {
      "members": [
        "you"
      ],
      "state_name": {
        "2023": "AK"
      }
    }
  }
}
//...
```

### DIFF
--- a/src/api/variables.js
+++ b/src/api/variables.js
@@ -509,7 +509,7 @@ export function optimiseHousehold(household, metadata, removeEmpty = false) {
           household[entityPlural][entityName][variable]
         )) {
           let variableData = newHousehold[entityPlural][entityName][variable];
-          if (variable === "members") {
+          if (variable === "members" || variable === "state_name") {
             continue;
           }
           let defaultValue = metadata.variables[variable].defaultValue;
--- a/src/api/variables.js
+++ b/src/api/variables.js
@@ -509,7 +509,7 @@ export function optimiseHousehold(household, metadata, removeEmpty = false) {
           household[entityPlural][entityName][variable]
         )) {
           let variableData = newHousehold[entityPlural][entityName][variable];
-          if (variable === "members" || variable === "state_name") {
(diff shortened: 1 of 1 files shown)


nikhilwoodruff (commented):
Thanks- yes this works! But I think a slightly better check (non-US specific) might be to exempt if it's in `metadata.basicInputs`? LMK if you run into any issues there though, thanks!

samau3:
> exempt if it's in `metadata.basicInputs`? 

Made the change to use `metadata.basicInputs`! Just wanted to note that this causes the `situation` to include the default variable values, in case that's a concern.

nikhilwoodruff (approved):
Thanks @samau3 - all LGTM now!

Issue #78, opened 2022-12-17
Replace baseline+reform vs diff switch with a radio button component


MaxGhenis:
Let's do a [tab-style radio button](https://ant.design/components/radio#components-radio-demo-radiobutton) here.

<img width="761" alt="image" src="https://user-images.githubusercontent.com/6076111/208252697-4dc5b4fe-b8a4-432b-bb40-5f13a8f43f0b.png">

On both the vary earnings and MTR charts

MaxGhenis:
Reopening to do this for the MTR chart as well

PR #327, opened 2023-03-15
Replace baseline+net vs diff switch with a radio button component #78

nung22:


MaxGhenis:
Thanks @nung22! Could you please add a screenshot and/or screen recording of it in action?

nung22:
policy-engine-issue-78
Here's a gif of the new tab-style radio button!

nikhilwoodruff:
Thanks @nung22! This is helpful, OK so two minor requests:
* Could we use the same blue as the rest of the app (and the chart)? There's a `const BLUE = ...` in the repo.
* Could we also make sure the padding between the bottom of the radio button component and the start of the chart legends is the same so the chart isn't jumping up and down too much?

MaxGhenis:
Could you remove `Show` from the radio button labels?

nikhilwoodruff (approved):

nung22:
v3
@MaxGhenis Here it is with `Show` removed from radio button labels

nikhilwoodruff:
Thanks @nung22. Merging now

PR #334, opened 2023-03-17
Drop modify_parameters and reform if no reform

### LINKED ISSUE #115: Drop reform bits of household `Reproduce in Python` when no reform provided
Or if not, fill in the `modify_parameters` function to be valid Python
image

samau3:
Fix #115.

If no reform is provided, this change removes the `modify_parameters(parameters)` line from the Python block and returns the function before adding the strings for `class reform(Reform)`. 

image

### DIFF
--- a/src/pages/policy/output/PolicyReproducibility.jsx
+++ b/src/pages/policy/output/PolicyReproducibility.jsx
@@ -69,7 +69,8 @@ export function getReformDefinitionCode(metadata, policy) {
   ];
 
   if (Object.keys(policy.reform.data).length === 0) {
-    lines.push("    pass");
+    lines.pop();
+    return lines;
   }
 
   for (const [parameterName, parameter] of Object.entries(policy.reform.data)) {


nikhilwoodruff:
Thanks this is great- though could we check it definitely produces correct Python syntax? Just cause it looks like `reform` might be an undefined variable (we should skip that line I think).

samau3:
Apologies for not catching that! Had to make a change similar to what was done for the `PolicyReproducibility` file in `HouseholdReproducibility` by utilizing `Object.keys(policy.reform.data)`. 

Previously, there was a ternary to determine if the `reform` variable would be included in the Python block relying on `policy.data.reform`. However, since an empty object is still truthy, this ternary would not properly determine if the `reform` variable should be included in the final string. 

image

nikhilwoodruff (approved):

nikhilwoodruff:
Thanks @samau3 - merging!

PR #344, opened 2023-03-19
Issue #318: Use different colors for credits and benefits vs taxes

### LINKED ISSUE #276: Remove `Your benefits don't change` section when a policy reform doesn't change them
[Example](https://policyengine.org/us/household?focus=householdOutput.netIncome&household=7706&region=us&timePeriod=2023&baseline=2&reform=4180)

image

### LINKED ISSUE #318: Use different color for tax on net income breakdown screen
e.g. red

Thanks to @xudonglai0426 for the suggestion!

PGrad:
Fixes #318 and fixes #276.

### Color scheme for additions or subtractions to net income (Issue #318)
```
src/
    - pages/household/output/NetIncomeBreakdown.jsx
    - style/colors.jsx
```
PolicyEngineArrows

The color scheme I'm currently using is dark green for anything that increases your net income, like benefits or tax credits, and red for decreases to net income like taxes. This scheme is also applied when you're evaluating a policy's effect on net income. Note that I also use up and down arrows to convey the same information for [red/green colorblind](https://www.nei.nih.gov/learn-about-eye-health/eye-conditions-and-diseases/color-blindness/types-color-blindness) users; I initially tried using line styles but it made the text harder to read.

The net income breakdown tree's logic is controlled by the `VariableArithmetic` component, which separates child nodes into additions or subtractions. I can apply color based on whether the node is adding or subtracting, so I'm not actually checking "is this a tax credit?" or "is this a benefit?" at every node. Of course, when a user applies a policy onto the net income breakdown, things get a little more complex; for example, a decrease to a tax is an increase to net income, so it has to be green and not red.

I added a `DARK_RED` color to colors.jsx, but if we want to use dark grey instead we can do that too.

The item descriptions have been moved into tooltips:
Tooltip

### Remove plus icon when your benefits/taxes/whatever don't change (Issue #276)
```
src/pages/household/output/NetIncomeBreakdown.jsx
```
If at a node it calculates that there is no change, I remove the plus icon. Before you would still see the plus icon and could expand the tree, and there would be no child nodes. You can see I removed the plus icon in the image above, at the node that says "Your employee-side payroll tax doesn't change".

### Fixed typo
```
src/controls/InputField.jsx
```
I saw an error that `inputmode` should be `inputMode` on the `motion.input` component.


src/controls/InputField.jsx
```diff
@@ -28,7 +28,7 @@ export default function InputField(props) {
         marginTop: mobile ? 0 : -10,
       }}
       type={type || "tel"}
-      inputmode={inputmode || "decimal"}
+      inputMode={inputmode || "decimal"}
```
nikhilwoodruff, line 31: Thank you, good spot- will be good to remove that warning from the console.

src/pages/household/output/NetIncomeBreakdown.jsx
```diff
@@ -8,16 +9,66 @@ import {
...
+        fontSize: "1.5em",
+        color: style.colors.DARK_GREEN,
+      }}
+    >
+      &uarr;
+    </span>
+  );
+}
+
+function DownArrow() {
```
nikhilwoodruff, line 29: I like this UI feature, nice- could we adjust the alignment slightly though? So it's inline and (if not non-trivial) same-height with the text: currently looks like it's vertically off a bit.

nikhilwoodruff (commented):
This is really nice, thanks @PGrad. @MaxGhenis I like the red here and think it's OK given the arrows solve the colour blindness issue, WDYT? If all agreed here I think we can merge after the slight alignment improvement on the arrows.

PGrad:
@nikhilwoodruff I agree. I was trying to fix the arrow alignment but I wasn't quite sure how to do it. @martinyis do you have an idea on how to fix it? The relevant code is here at line 12 in NetIncomeBreakdown.jsx:
```ts
// The arrows are used to differentiate increases or decreases
// to net income for colorblind users.
function UpArrow() {
  return (
    <span
      style={{
        display: "inline-block",
        fontWeight: "bold",
        fontSize: "1.5em",
        color: style.colors.DARK_GREEN,
      }}
    >
      &uarr;
    </span>
  );
}

function DownArrow() {
  return (
    <span
      style={{
        display: "inline-block",
        fontWeight: "bold",
        fontSize: "1.5em",
        color: style.colors.DARK_RED,
      }}
    >
      &darr;
    </span>
  );
}
```

nikhilwoodruff:
@PGrad thanks.

> I'm also noticing my css is a little repetitive here - I could write a SASS class for these arrows maybe.

What are your thoughts on SASS classes vs storing CSS objects/constants in a JS file/set of files? I actually find the latter easier given it keeps more things consistently in JS, but I'm not sure I've seen it much elsewhere.

martinyis:
> @nikhilwoodruff I agree. I was trying to fix the arrow alignment but I wasn't quite sure how to do it. @martinyis do you have an idea on how to fix it? The relevant code is here at line 12 in NetIncomeBreakdown.jsx:
> 
> ```ts
> // The arrows are used to differentiate increases or decreases
> // to net income for colorblind users.
> function UpArrow() {
>   return (
>     <span
>       style={{
>         display: "inline-block",
>         fontWeight: "bold",
>         fontSize: "1.5em",
>         color: style.colors.DARK_GREEN,
>       }}
>     >
>       &uarr;
>     </span>
>   );
> }
> 
> function DownArrow() {
>   return (
>     <span
>       style={{
>         display: "inline-block",
>         fontWeight: "bold",
>         fontSize: "1.5em",
>         color: style.colors.DARK_RED,
>       }}
>     >
>       &darr;
>     </span>
>   );
> }
> ```
> 
> I'm also noticing my css is a little repetitive here - I could write a SASS class for these arrows maybe.

I would use display: "inline", instead of display-blockif you want to fix alignment but in order for me to see the whole problem I need to know how to actually to see NetIncomeBreakdown.jsx in the browser. I started server but I am a bit lost how to navigate to this page

PGrad:
I fixed the alignment issue:
AlignmentFix

I fix this problem by using `inline-flex`. A side effect of this choice (which I think we can live with) is that when the viewport width is squished, the arrow and dollar amount get separated from the text in the child nodes of the tree:
AlignmentFixSquished

PGrad:
> What are your thoughts on SASS classes vs storing CSS objects/constants in a JS file/set of files? I actually find the latter easier given it keeps more things consistently in JS, but I'm not sure I've seen it much elsewhere.

I think either is fine, as long as the CSS is kept DRY.

PGrad:
@nikhilwoodruff  Now the colors are green and grey:
green_and_grey

nikhilwoodruff:
Yeah this is way better, love it.

PR #346, opened 2023-03-20
Fix issue 342: Cliff shading height

### LINKED ISSUE #342: Cliff shading only goes up to $200k tall
[Example](https://policyengine.org/us/household?focus=householdOutput.earnings&household=16193)

image

samau3:
Fixes #342.

Previously, the height of the cliff shading was hardcoded to go up to `200,000`. This PR updates the max shading height to be the ceiling of the net income array rounded up to the nearest `100,000`.

household=16193


nikhilwoodruff (approved):
Looks great, thanks @samau3!

PR #350, opened 2023-03-21
Fix Issue 341: Display changes to MTR when there are changes

### LINKED ISSUE #341: MTR diff chart says no change when there is a change
[Example](https://policyengine.org/us/household?focus=householdOutput.mtr&reform=7124&region=us&timePeriod=2023&baseline=2&household=15238)

image

samau3:
Fixes #341.

Previously the comparison made to determine if a reform impacted MTR was made using `baselineMtrValue` and `currentMtr`. This caused a disconnect between what the charts show for the effects of a reform on MTR and what the descriptions of the impact on the page. 

With this PR, the comparison made is with the reform MTR value that corresponds to the current earnings amount. 

MTR chart

MTR difference

### DIFF
--- a/src/api/variables.js
+++ b/src/api/variables.js
@@ -526,3 +526,21 @@ export function optimiseHousehold(household, metadata, removeEmpty = false) {
   }
   return newHousehold;
 }
+
+export function findEarningsIndex(earningsArray, currentEarnings) {
+  let left = 0;
+  let right = earningsArray.length - 1;
+
+  while (left <= right) {
+    let middle = Math.floor(left + right) / 2;
+    if (earningsArray[middle] === currentEarnings) {
+      return middle;
+    }
+    if (earningsArray[middle] > currentEarnings) {
+      right = middle - 1;
+    } else {
+      left = middle + 1;
+    }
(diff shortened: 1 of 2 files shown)


nikhilwoodruff:
Thanks, just from a quick look- do we need the binary search function, given we already have the value from the 'you are here' marker?

samau3:
>  we already have the value from the 'you are here' marker

The `you are here` marker value represents the `currentEarnings` value which isn't enough information to find which reform MTR element it corresponds to in the `reformMtr` array. For example, the `you are here` marker in the issue is `500000` but that doesn't translate to which `reformMtr` element corresponds to that income amount as it's elements range from `0` to `0.5625`.

Since the `earningsArray` lines up with the `reformMtr` array, finding the index that `currentEarnings` appears in the `earningsArray` could be used to find the `reformMtr` element at that earnings amount. I originally used `indexOf(currentEarnings)`, but since the `earningsArray` is sorted in order, I opted to utilize binary search.

nikhilwoodruff (approved):
Thanks @samau3!

PR #357, opened 2023-03-23
Maintain JSON formatting for situation in Python block

### LINKED ISSUE #167: Apply indented line breaks to `Reproduce in Python` household dict
It's currently one super-long line

samau3:
Fixes #167.

The JSON formatting that `householdJson` already has from being returned from `JSON.stringify()` was being reformatted by the browser. By using `<pre>`, the `JSON` styling (e.g. the new lines and white space) is kept when being displayed on the browser. 

Note, this causes the python block to be larger than before. 

JSON Formatted Situation

### DIFF
--- a/src/pages/household/output/HouseholdReproducibility.jsx
+++ b/src/pages/household/output/HouseholdReproducibility.jsx
@@ -37,7 +37,21 @@ function PythonCodeBlock({ lines }) {
         {lines.map((line, i) => {
           if (line === "") {
             return <div key={i} style={{ paddingTop: 15 }} />;
-          } else {
+          } else if (line.includes("situation = {")) {
+            return (
+              <pre 
+                key={i} 
+                style={{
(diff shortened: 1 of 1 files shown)


nikhilwoodruff (commented):
Thanks @samau3 - OK this looks great; my only request is could we move the `Copy` button to above the code block now, given users might have to scroll down a fair bit to see it with the increased line count? Good to merge after that I think.

samau3:
> could we move the `Copy` button to above the code block now, given users might have to scroll down a fair bit to see it with the increased line count?

Good point about making the `Copy` button more visible when there's a large line count! I opted to restrict the max height of the Python block to be `50vh` and scrollable so the `Copy` button placement is consistent across both Household and Policy. 

Scrollable Python Block

nikhilwoodruff:
Ooh yes that's a better idea. Thanks- merging!

Issue #125, opened 2022-12-29
Prevent empty hover labels


nikhilwoodruff:
Pretty straightforward - just needs a check of the hover event data to avoid setting the hovercard to a non-null value in all the policy impact chart components.

nikhilwoodruff:
@jacky5124 to take a start on this

jacky5124:
Was fixed by pull request #365

PR #365, opened 2023-03-23
Prevent hover cards from displaying when data points are unhovered

jacky5124:
Fixes issue #125 

Every plot that uses HoverCard did not clear the content of the hover card when the mouse was moved away from data points, hence the hover card was always displaying the last data point the mouse was hovered over. Now this is fixed.

HoverCard component kept rendering the hover card even though the content was empty. Now a check is added to the rendering logic so that the hover card will not appear whenever the content is empty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ yarn-error.log*
 /requirements.txt
 **/__pycache__
 .env
+
+.idea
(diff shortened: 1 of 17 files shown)


nikhilwoodruff (approved):
Thanks @jacky5124! This looks great and I confirmed it works locally. Apologies that the existing code has some duplication :). 

Merging now- in future could we please add a screenshot or a GIF demonstrating in the PR so others can easily see how it changes the app without needing to check it out locally? Thanks!

jacky5124:
Thank you! I will include screenshots or git the next time I work on the
app repo.

Issue #371, opened 2023-03-25
Make links in blog posts open in a new tab


nikhilwoodruff:

shyamtawli:
I can work on it. Please assign or should I start working on it.

nikhilwoodruff:
Thanks @shyamtawli! Assigned you

PR #372, opened 2023-03-26
issue:371 blog post open in new tab

shyamtawli:
This PR will resolve Issue #371 

I have removed `useNavigate` and used `window.open` method to open blogs on new tab when user clicks on it.

### DIFF
--- a/src/layout/BlogPostHolder.jsx
+++ b/src/layout/BlogPostHolder.jsx
@@ -2,7 +2,6 @@ import postJson from "../posts/posts.json";
 import authorsJson from "../posts/authors.json";
 import style from "../style";
 import { motion } from "framer-motion";
-import { useNavigate } from "react-router-dom";
 import EmailSignUp from "./EmailSignup";
 import moment from "moment/moment";
 
@@ -18,7 +17,6 @@ function BlogPostPreviewRegular(props) {
     name = name.substring(3);
   }
   const imageSrc = require(`../images/posts/${image}`);
-  const navigate = useNavigate();
   const authorImages = authors.map((author) => {
     return require(`../images/authors/${authorsJson[author].headshot}`);
   });
@@ -39,7 +37,7 @@ function BlogPostPreviewRegular(props) {
         cursor: "pointer",
       }}
       whileHover={{ scale: 1.05 }}
-      onClick={() => navigate(`/${countryId}/blog/${name}`)}
+      onClick={() => window.open(`/${countryId}/blog/${name}`, "_blank")}
     >
       <img
         src={imageSrc}


nikhilwoodruff:
Thanks @shyamtawli! Sorry- I should've been more explicit: I meant that the links inside a blog post should open in a new tab, rather than the links to the blog posts on the homepage.

I'm undecided on whether we want the links to the blog posts to open in a new tab, but leaning towards no. The important reason to want the internal blog post links to open in a new tab is to avoid closing the PolicyEngine tab if they lead to external sites, but this isn't the case with the links to blog posts.

nikhilwoodruff (changes requested):
Thanks @shyamtawli! Sorry- I should've been more explicit: I meant that the links inside a blog post should open in a new tab, rather than the links to the blog posts on the homepage.

I'm undecided on whether we want the links to the blog posts to open in a new tab, but leaning towards no. The important reason to want the internal blog post links to open in a new tab is to avoid closing the PolicyEngine tab if they lead to external sites, but this isn't the case with the links to blog posts.

shyamtawli:
Sorry I also misread the issue title. 
Let me clear now that the links which are on blog post should be open on new tab. right?
All the links that are attached on words should open on new tab?

nikhilwoodruff:
@shyamtawli yep- for example, the 'policy costings here' link should open in a new tab.

image

shyamtawli:
@nikhilwoodruff Okay, will implement it ASAP 👍

shyamtawli:
@nikhilwoodruff I noticed that all the blogs are rendered from a Markdown file. At first, I thought of adding `target="_blank"` to all the links, but this would not be very efficient.

After doing some research, I discovered that adding `<base target="_blank">` to the top of the Markdown file will open all the links in a new tab.

Do you know of any other ways to achieve this? Please let me know.

shyamtawli:
@nikhilwoodruff Can you share the file path where the `a` tag renders

nikhilwoodruff:
Yep 

> @nikhilwoodruff Can you share the file path where the `a` tag renders
> 
> Update: I have tried my way and it works perfectly fine.

Yep, here https://github.com/PolicyEngine/policyengine-app/blob/2b9a366be8ba5adaca3834baf94c347ca16b2243/src/pages/BlogPage.jsx#L374

shyamtawli:
@nikhilwoodruff I have made the chances and it is fixed by one line of code.

Should I fix the lint errors or it is fine?

nikhilwoodruff (approved):
Perfect, thanks! The lint check seems to be passing, so I'll merge this now.

shyamtawli:
@nikhilwoodruff Thanks for guiding me. 
I will see other issues as well and will try to resolve it.

PR #375, opened 2023-03-26
Personalize GPT analysis with ELI5 and Wonk modes, and rename to Analysis

MaxGhenis:
Still want to check some of the results and maybe tweak the prompt more before merging

### DIFF
--- a/src/pages/policy/output/Analysis.jsx
+++ b/src/pages/policy/output/Analysis.jsx
@@ -131,13 +131,10 @@ export default function Analysis(props) {
     <>
       <h2>Analysis</h2>
       <p>
-        PolicyEngine&apos;s automatic policy analysis is powered by GPT-4. Read
-        more about this feature{" "}
         <a href="https://policyengine.org/uk/blog/2023-03-17-automate-policy-analysis-with-policy-engines-new-chatgpt-integration">
-          here
+          Read more about PolicyEngine&apos;s automatic GPT4-powered policy analysis.
         </a>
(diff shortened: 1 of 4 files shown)


src/pages/policy/output/Analysis.jsx
```diff
@@ -59,14 +70,44 @@ export default function Prompt(props) {
...
+    setAudience(audienceValue);
+  };
+
+
+  const buttonWidth = "80px";
+  const activeColor = "#6c757d";
+  const inactiveColor = "white";
+  const borderColor = "1px solid #6c757d";
+
+  const AudienceButton = ({ audienceValue, currentAudience, handleAudienceChange }) => (
```
nikhilwoodruff, line 85: We do use functional components in the codebase, but all the others use the same `function AudienceButton() {}` syntax.

src/pages/policy/output/Analysis.jsx
```diff
@@ -59,14 +70,44 @@ export default function Prompt(props) {
...
-  const lines = analysisPrompt.split("\n");
+  const lines = prompt.split("\n");
+
+  const handleAudienceChange = (audienceValue) => {
+    setAudience(audienceValue);
+  };
+
+
+  const buttonWidth = "80px";
+  const activeColor = "#6c757d";
```
nikhilwoodruff, line 81: If this is `DARK_GRAY` we have a `style.colors` constant holding it, would be good to link it up to that.

nikhilwoodruff (approved):

PR #403, opened 2023-03-28
Issue 360: adds social share links to policy and household page

QianruZhang78:


QianruZhang78:
Screen Shot 2023-03-28 at 6 22 11 PM
household page
Screen Shot 2023-03-28 at 6 22 30 PM
copy link alarm
Screen Shot 2023-03-28 at 6 23 00 PM
policy page
Screen Shot 2023-03-28 at 6 23 32 PM
mobile

QianruZhang78:
@nikhilwoodruff The pull request is ready for review, however I am not sure if it is possible to pre-populate the title into twitter post like before. If not I guess we can pre-populate a tag such as #PolicyEngine etc so we can track the influence that way. Let me know what u think about it:)

nikhilwoodruff (changes requested):
This looks great @QianruZhang78! Hope you don't mind, I played about with a few variations on your changes locally and made some minor styling changes (you should be able to get them if you `git pull` on your machine). For me it now looks like:

image

Could we just:

- [x] Check that my changes haven't thrown off anything on the other pages (nothing looks obviously wrong when navigating about). I took out some padding from the other components in order to shift the share links component upwards a bit, which might have had some side effects I missed.
- [x] Also apply the same styling to the household output page
- [x] For the title, could we use "{policyLabel}, on PolicyEngine"? The policy label should be in the component as a variable you can access.
- [ ] Also just check that this all looks OK on mobile

Then after these I think we can merge this!

nikhilwoodruff:
Oh also one other thing: could we use [this ant-design component](https://ant.design/components/message) instead of `alert` for the link copy status message? Thanks!

QianruZhang78 (commented):
@nikhilwoodruff I have completed the check list and this is what it looks like in the household page now(with link copy status message updated):

Screen Shot 2023-04-01 at 12 54 05 AM

This is how it looks like in the policy page:

Screen Shot 2023-04-01 at 12 46 18 AM

And this is how the post with title look like:

Screen Shot 2023-04-01 at 12 22 12 AM

I don't think there's any side effects of the styling changes u made neither, and these all looks OK on mobile. Please check:)

nikhilwoodruff:
Perfect, no notes- thanks @QianruZhang78 this looks amazing!

nikhilwoodruff (approved):

PR #423, opened 2023-04-01
Add documentation embed

nikhilwoodruff:
Thought we could do a simple documentation integration


MaxGhenis (approved):
Looks great. Could you use a US screenshot for the US?

Also how's this (GPT-suggested) text instead? The current text risks implying you have to use the package to do simulations, and country package is ambiguous IMO.

header: Advanced analysis with our Python packages

text: Dive deeper into policy impact analysis using PolicyEngine's open-source Python packages. Customize your simulations and perform advanced reforms for thorough insights, all on your own computer.

link: Explore the documentation →

Issue #231, opened 2023-02-10
Remove date from post URLs


MaxGhenis:
Blogs seem to be moving away from the yyyy-mm-dd URL prefix

It'd be nice but not necessary to redirect the old URLs

jacky5124:
Was fixed by pull request #439

PR #439, opened 2023-04-05
Remove YYYY-MM-DD from blog URLs

jacky5124:
Fixes issue #231

All markdown filenames for blogs are renamed so that YYYY-MM-DD prefixes are no longer there.

A new script called rename.sh under src/posts folder which can automate all the renaming has been added, and the README under src/posts which instructs how to create blog markdown files has also been updated.

Old URLs, which have the prefixes, will be automatically redirected to their corresponding new URLs.

Screen Recording 2023-04-05 at 4 03 13 PM

In addition, blog URLs which never existed will be redirected to the 404 error page.

Screen Recording 2023-04-05 at 4 06 55 PM


nikhilwoodruff (approved):
Amazing, thanks for this! Looks like it works perfectly (tested a few examples locally).

PR #454, opened 2023-04-12
Add social cards for economic impacts

### LINKED ISSUE #356: Social cards for economic impact charts
This is a probably technically complex (or not, maybe there's an easy way to do it) task. But the social card for each chart page could be a screenshot of that titled chart. This would be massively useful on Twitter, where we get most of our engagement (it would probably boost that by a fair amount: currently, we put up screenshots and bury the PolicyEngine link in a later tweet in a thread- with this, we could post a screenshot and a link all in one).

I think the way we'd want to do this is:
* In the Flask app, check if the URL has a ?social=True query parameter
* If it does, render the page in a headless browser in Python[^1]
* Save the image somewhere (probably locally - can always regenerate them)
* Set the twitter social card metadata in the index.html tags, including the URL of the saved image

[^1]: Or make the image with some Python package. It seems easier to mirror the web app by taking a screenshot though

samau3:
This PR aims to fix #356.

Current progress adds the ability to take a screenshot of the chart area element on the page. 

But more changes are needed to:
- Send the image file to a backend route for hosting
- Receive an image URL from the backend
- Update the metatags of the page such that the proper image is displayed


nikhilwoodruff:
Just made a few tweaks to get the image saving working. It looks like it all works OK locally now! So I might merge this and test it out.

nikhilwoodruff:
Just hit me actually that I think Twitter caches social cards! So we might not even need to store them persistently.

PR #538, opened 2023-05-11
Update object used for `/calculate` on API Status page

samau3:
Since the expected data used in POST requests for `/calculate` changed following updates to the API packages for `US` and `CA`, the API status page encountered errors when making requests to those routes with the outdated objects. 

image

This led to inaccurate status messages for `US` and `CA` despite the API routes for those countries' `/calculate` being responsive. With this PR, the `US` and `CA` objects used when making requests to `/calculate` are updated to reflect the new object structure expected for their respective routes' `POST` requests.


nikhilwoodruff (approved):
Thanks @samau3!

Issue #501, opened 2023-04-28
Break down US budgetary impact by state and federal taxes


MaxGhenis:
For example, most analysts of a state level income tax reform will focus on the impact to state budgets, but we currently include federal budget impacts (via the SALT deduction) as well. Let's separate these effects in the waterfall chart.

PR #542, opened 2023-05-17
Fix issue #501: distinguish State vs federal budgetary impacts

QianruZhang78:
### Summary
🌎💰📊

This pull request enhances the `BudgetaryImpact` component to support regional policy simulations. It uses the URL parameter to fetch and display the appropriate budgetary impact data for each region.

> _To simulate policies in regions_
> _We updated the `BudgetaryImpact` legions_
> _We got the param from the URL_
> _And tweaked the chart to show it well_
> _And also removed some code redundancies_

### Walkthrough
*  Add code to get the region parameter from the URL query string in `BudgetaryImpact.jsx` ([link](https://github.com/PolicyEngine/policyengine-app/pull/542/files?diff=unified&w=0#diff-af88a6bb0672282e83959a0a8317e5259f5e148af72f4901f431073ca751e7b3R19-R21))
*  Modify the values and labels for the budgetary impact chart based on the region parameter in `BudgetaryImpact.jsx` ([link](https://github.com/PolicyEngine/policyengine-app/pull/542/files?diff=unified&w=0#diff-af88a6bb0672282e83959a0a8317e5259f5e148af72f4901f431073ca751e7b3L26-R45))
*  Remove the redundant code to get the region parameter in the `BudgetaryImpactChart` component in `BudgetaryImpact.jsx` ([link](https://github.com/PolicyEngine/policyengine-app/pull/542/files?diff=unified&w=0#diff-af88a6bb0672282e83959a0a8317e5259f5e148af72f4901f431073ca751e7b3L141-L142))

### DIFF
--- a/src/pages/policy/output/BudgetaryImpact.jsx
+++ b/src/pages/policy/output/BudgetaryImpact.jsx
@@ -19,30 +19,25 @@ export default function BudgetaryImpact(props) {
 
   const urlParams = new URLSearchParams(window.location.search);
   const region = urlParams.get("region");
-  const desktopLabels = [
+
+  let desktopLabels = [
     "Federal tax revenues",
     "State tax revenues",
     "Benefit spending",
     "Net impact",
   ];
-  const mobileLabels = ["Federal taxes", "State taxes", "Benefits", "Net"];
-  let valuesBeforeFilter, labelsBeforeFilter;
-  if (region === "us") {
-    valuesBeforeFilter = [
-      taxImpact / 1e9,
-      stateTaxImpact / 1e9,
-      -spendingImpact / 1e9,
-      budgetaryImpact / 1e9,
-    ];
-    labelsBeforeFilter = mobile ? mobileLabels : desktopLabels
-  } else {
-    valuesBeforeFilter = [
-      taxImpact / 1e9,
-      stateTaxImpact / 1e9,
-      -spendingImpact / 1e9,
-    ];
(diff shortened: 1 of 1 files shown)


MaxGhenis:
@nikhilwoodruff thought you did this already?

nikhilwoodruff:
Yes, sorry @QianruZhang78 I'm not sure why the issue wasn't closed - my bad here. That said, I think this is an improvement still within the code. There was one bit I missed out when I originally implemented this logic: outside the US and Canada, we don't really call it 'federal tax revenues' but just 'tax revenues'. Could we add a bit of logic that applies this shorter name where the countryId is not us or ca? Thanks!

QianruZhang78:
I've add the logic that will apply the shorter name when the countryId is not us or ca. However when I try to test the case for Nigeria and Canada this error pop up, not sure if this is something we already aware of, for details see pictures below.

uk:
Screen Shot 2023-05-20 at 11 35 31 PM
us:
Screen Shot 2023-05-20 at 11 34 29 PM
error:
Screen Shot 2023-05-20 at 11 30 06 PM

nikhilwoodruff (approved):
Perfect- and thanks for identifying. This is actually totally fine because we don't yet support microsimulation in Canada or Nigeria. I suppose neatening up the edges there to make sure the whole interface isn't visible at all could be a nice improvement but I think we can definitely merge this. Thanks @QianruZhang78!

PR #555, opened 2023-05-26
feat: replace create-react-app and add vite

### LINKED ISSUE #471: Convert create-react-app/webpack to Vite
Updating to Vite will dramatically decrease local dev times and make for a much more pleasant devEx.

eugenedwk:
Fixes #471 

Create-react-app has an infamously slow spinup time among other on-going issues. Vite is a more recently popular tooling that uses ESModules among other modern tech that speeds up the build process.

Time to spin up local environment originally clocked in at 1:30.97 reduced to 3.722.


src/pages/BlogPage.jsx
```diff
@@ -183,7 +163,7 @@ export function BlogPostMarkdown(props) {
           <td
             style={{
               padding: 5,
-              fontFamily: "Roboto Serif",
+              fontFamily: "Merriweather",
```
MaxGhenis, line 166: Intended?

eugenedwk: I don't remember the exact reason but there was some quirk about the roboto serif font that prevented the transition. If there's a hard brand rule we're bumping up against I can put roboto back in and see what the errors were.

eugenedwk: I don't remember the exact reason but there was some quirk about the roboto serif font that prevented the transition. If there's a hard brand rule we're bumping up against I can put roboto back in and see what the errors were.

MaxGhenis:
Can we move this forward or close it? Curious how it fits with the redesign @anth-volk @nikhilwoodruff

nikhilwoodruff:
Sorry for my delay here. Looks OK to me but let's revisit after the redesign completes just for safety.

eugenedwk:
Sorry for the delay on my part as well. There was a build error when I had submitted this PR because of some mixed use of import/require that vite didn't like. Happy to revisit after the website rebuild to see if we can resolve those errors because I strongly believe vite or any other non-CRA tooling makes development so much more pleasant.

anth-volk:
@eugenedwk Apologies for the delay in getting to this, our focus turned to higher-priority fixes and this fell down the list. Would you still be willing to address the build issues, as we've now completed the redesign?

eugenedwk:
> @eugenedwk Apologies for the delay in getting to this, our focus turned to higher-priority fixes and this fell down the list. Would you still be willing to address the build issues, as we've now completed the redesign?

No worries! totally makes sense and congrats on the design overhaul. Given the extensive changes I'll close this PR and resubmit based on the current repo instead of trying to rebase.

anth-volk:
Thanks so much, that'd be very appreciated.

PR #560, opened 2023-05-29
Fix #459: Add caution modal for US impacts of State tax reforms

### LINKED ISSUE #459: Add caution modal for US impacts of State tax reforms
Sometimes users edit a state tax parameter, but then simulate over the US (which freezes state taxes to avoid biasing in favour of those we've implemented). What we could do is:
* When the user presses the "calculate economic impact" button, check if they've got both state tax reforms and a region of "us"
* Open an "are you sure?" dialog to check this is intended.

QianruZhang78:
### Summary
🛠️💬🚀

This pull request adds a confirmation dialog for calculating the economic impact of state tax reforms on the policy page. It also enhances the `NavigationButton` component to allow custom click handlers. These changes improve the user experience and functionality of the policy page.

> _To make `NavigationButton` more clever_
> _We added an `onClick` prop lever_
> _It can show a `Modal`_
> _Before changing the total_
> _Of the state tax reform's endeavor_

### Walkthrough
*  Add confirmation dialog for economic impact calculation ([link](https://github.com/PolicyEngine/policyengine-app/pull/560/files?diff=unified&w=0#diff-c75e6033e75a045bd80335e28a44c550877d3ea9388a54de8b7955e9b0508285L15-R15), [link](https://github.com/PolicyEngine/policyengine-app/pull/560/files?diff=unified&w=0#diff-c75e6033e75a045bd80335e28a44c550877d3ea9388a54de8b7955e9b0508285R203-R220), [link](https://github.com/PolicyEngine/policyengine-app/pull/560/files?diff=unified&w=0#diff-c75e6033e75a045bd80335e28a44c550877d3ea9388a54de8b7955e9b0508285L357-R375))
*  Allow custom `onClick` prop for `NavigationButton` component ([link](https://github.com/PolicyEngine/policyengine-app/pull/560/files?diff=unified&w=0#diff-fc3284b09378d8360c8bcfe2e7e44d6cbe6c4e806becdf3ace25facd8bb619a1L7-R36))

### DIFF
--- a/src/pages/policy/PolicyRightSidebar.jsx
+++ b/src/pages/policy/PolicyRightSidebar.jsx
@@ -13,6 +13,7 @@ import style from "../../style";
 import { RegionSelector, TimePeriodSelector } from "./output/PolicyOutput";
 import PolicySearch from "./PolicySearch";
 import { Alert, Modal} from "antd";
+import { ExclamationCircleOutlined } from "@ant-design/icons";
 
 function PolicyNamer(props) {
   const { policy, metadata } = props;
@@ -198,17 +199,36 @@ export default function PolicyRightSidebar(props) {
   const reformPolicyId = searchParams.get("reform");
   const baselinePolicyId = searchParams.get("baseline");
   const focus = searchParams.get("focus") || "";
+  const stateAbbreviation = focus.split(".")[2];
   const hasHousehold = searchParams.get("household") !== null;
   const [showReformSearch, setShowReformSearch] = useState(false);
+  const options = metadata.economy_options.region.map((stateAbbreviation) => {
+    return { value: stateAbbreviation.name, label: stateAbbreviation.label };
+  });
+  const label = options.find((option) => option.value === stateAbbreviation)?.label;
   const confirmEconomicImpact = () => {
     if (region === "us" && focus.startsWith("gov.states")) {
       Modal.confirm({
-        title: "Confirm",
-        content: "You are about to calculate the economic impact of a state tax reform for the entire US. Are you sure you want to proceed?",
+        title: " You are about to calculate the economic impact of a state tax reform for the entire US",
+        content: `To compute economic impacts over ${label} only, select that option in the region selector on the right pane.
+         Are you sure you want to proceed?`,
+        icon: <ExclamationCircleOutlined style={{ color: "#2C6496" }} />,
         onOk() {
           let newSearch = copySearchParams(searchParams);
           newSearch.set("focus", "policyOutput");
           setSearchParams(newSearch);
+        },
+        okButtonProps: {
+          style: {
+            backgroundColor: "#2C6496", 
+            borderColor: "#2C6496", 
+          },
+        },
+        bodyStyle: {
+          paddingLeft: '9px' 
+        },
+        style: {
+          top: "25%", 
         },
       });
     } else {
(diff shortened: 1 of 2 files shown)


QianruZhang78:
Screen Shot 2023-05-29 at 10 24 01 PM

As shown in the screen shot, when region is "us" and focus starts with "gov.states", this confirmation modal will pop up. Let me know if there's anything should be updated:)

nikhilwoodruff (changes requested):
Thanks @QianruZhang78! Logic-wise, I think this is good to go. Could we just make some minor edits to the styling and content?
- [x] Making the buttons the same style as the others (including using PolicyEngine's blue)
- [x] Adding a suggestion for how to actually change the region ("To compute economic impacts over [State] only, select that option in the region selector on the right pane").
- [x] Aligning the text body with the icon so there isn't a big gap on the left

Thanks!

QianruZhang78:
Screen Shot 2023-05-31 at 3 47 49 PM
@nikhilwoodruff Better this way?

QianruZhang78:
@nikhilwoodruff This thought just popped into my mind, what about using a different message when user trying to simulate a state tax reform over the wrong state? With the template now I suppose this can be done easily, I can add this update if you think it is appropriate:)

nikhilwoodruff:
@QianruZhang78 - looks much better and that's a great idea! Please do, thanks.

QianruZhang78:
> @QianruZhang78 - looks much better and that's a great idea! Please do, thanks.

Just an update, in the last #app meeting Max, Evan and Emily has provided some suggestions, where from an ux aspect, maybe it is better to have a "change state" button and a "continue" button, rather than just reminding users that they may have done something wrong. I really think that makes more sense and I'll try to make the update today:)

QianruZhang78:
Screen Shot 2023-06-02 at 12 20 08 AM
Screen Shot 2023-06-02 at 12 19 50 AM

After user clicking the "Change State" button, the page will reload with the updated region(I tried redirecting but couldn't make the region selector in the right side bar to update at the same time). LMK if there's anything should be updated, especially the phrasing, feels like there's gotta be something better than what I have now. @nikhilwoodruff

src/pages/policy/PolicyRightSidebar.jsx
```diff
@@ -198,8 +199,56 @@ export default function PolicyRightSidebar(props) {
   const reformPolicyId = searchParams.get("reform");
   const baselinePolicyId = searchParams.get("baseline");
   const focus = searchParams.get("focus") || "";
+  const stateAbbreviation = focus.split(".")[2];
```
nikhilwoodruff, line 202: This is a nice way of doing it. I guess it might not capture everything (e.g. if someone builds a reform including parameter changes at the state and federal level, and they're currently looking at a federal parameter) but this solves the immediate and most important use-case for sure.

src/pages/policy/PolicyRightSidebar.jsx
```diff
@@ -198,8 +199,56 @@ export default function PolicyRightSidebar(props) {
...
+  options.push({ value: region, label: region });
+  const label = options.find((option) => option.value === stateAbbreviation)?.label;
+  const regionLabel = options.find((option) => option.value === region)?.label;
+  const confirmEconomicImpact = () => {
+    let message = "";
+    if (stateAbbreviation && stateAbbreviation !== region) {
+      message = `You are about to calculate the economic impact of a tax reform in ${label} for ${regionLabel} `;
+    }
+    if (region === "us" && focus.startsWith("gov.states")) {
+      message = "You are about to calculate the economic impact of a state tax reform for the entire US ";
```
nikhilwoodruff, line 217:
```suggestion
      message = "You are about to calculate the economic impact of a state tax reform for the entire US, which PolicyEngine does not currently support. ";
```

nikhilwoodruff (approved):
@QianruZhang78 - this looks great. Added one minor change, but LGTM!

PR #587, opened 2023-06-21
Fixed the "shaking" when hovering charts and improved HoverCard performance

### LINKED ISSUE #504: Net income chart "shakes" when providing a reform
This has happened to me for a while, and I thought it was just my browser but someone else just reported the same issue.

The issue starts at 1:45

https://www.loom.com/share/4c2954ab10564ea6866f634946eb03a0

jacky5124:
Fixes #504 

Was originally exploring the hovertemplate prop suggested by #307 but later found a better way, which allows the custom HoverCard component to be kept and also prevents the chart from re-rendering, thus fixing the "shaking" bug and making the hovering experience smooth. In addition, some optimization is made to the HoverCard rendering logic, including wrapping the mouse move handler by useEffect to prevent memory leak, and managing the position of the HoverCard directly by a state to prevent the element from appearing at the top-left corner of the screen in some times.

HoverCard performance BEFORE this fix:

Shaking

HoverCard performance AFTER this fix:

Normal

Previously, the content of the HoverCard was managed by a state declared by each chart component (parent of the Plot and the HoverCard), where the Plot set the state and the HoverCard read the state, so that a change of the content meant a change of the state, resulting the Plot being re-rendered every time the mouse moved a bit, which caused the "shaking".

Now, the aforementioned state is no longer declared by each chart component, but directly by the HoverCard, and the state setter is offered through React context. The HoverCard exports a context, each chart component imports that context, and each Plot rendering call is factored into a standalone component which uses that context to obtain the state setter provided by the HoverCard. Since the state lives directly in the HoverCard, when the state (content) changes only the HoverCard itself will re-render, so the charts will no longer "shake" and the HoverCard itself will move much smoother.

copilot:all

### DIFF
--- a/src/layout/HoverCard.jsx
+++ b/src/layout/HoverCard.jsx
@@ -1,29 +1,27 @@
 import { AnimatePresence, motion } from "framer-motion";
-import { useState } from "react";
+import { createContext, useEffect, useState, memo } from "react";
 import style from "../style";
 import Divider from "./Divider";
 
+export const HoverCardContext = createContext((obj) => obj);
+
 export default function HoverCard(props) {
   const { children } = props;
-  let { content } = props;
-  if (!content) {
-    content = {
-      title: "",
-      body: "",
-    };
-  }
-  const notEmpty = content.title.length !== 0 || content.body.length !== 0;
-
-  const [visible, setVisible] = useState(false);
+  const [enabled, setEnabled] = useState(false);
+  const [position, setPosition] = useState({left: 0, top: 0});
+  const [content, setContent] = useState({title: "", body: ""});
+  const title = content && content.title ? content.title : "";
+  const body = content && content.body ? content.body : "";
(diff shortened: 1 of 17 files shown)


MaxGhenis:
Looks great, thank you @jacky5124!

IIRC, two other main reasons we were considering hovertemplate were:
1. to "snap" the cursor to the chart
2. showing up on mobile

How tractable do those seem keeping HoverCard?

Regardless we should merge this if @nikhilwoodruff approves the code, as it's a huge gain for the moment. Just thinking about where to invest next.

MaxGhenis (approved):

jacky5124:
I just learned that plotly.js has click-related event handler props, so HoverCard may be able to support mobile devices. Not sure how to "snap" the cursor to the chart though, but I am thinking about letting plotly.js draw a vertical line and a horizontal line crossing the data point at the cursor.

PR #593, opened 2023-07-03
Fixes #592 Lowercase State variable in questionnaire and remove caveat

### LINKED ISSUE #592: Lowercase State variable in questionnaire and remove caveat about calculations
https://policyengine.org/us/household?focus=input.household.state_name

image

I don't think this is in policyengine-us

marcadrian-it:
Changed each variable.label in VariableEditor to lowercase, and removed the caveat.


nikhilwoodruff (approved):
Thanks @marcadrian-it !

MaxGhenis:
Verified - thanks!
image

PR #606, opened 2023-07-08
Vertically center breadcrumbs

yellohCrayon:
copilot:all

Give height to breadcrumbs so they can be vertically centered 
Screenshot 2023-07-08 at 6 11 35 PM


nikhilwoodruff (approved):
Thanks @yellohCrayon!

PR #673, opened 2023-08-09
Add `hovertemplate` for all plots

### LINKED ISSUE #603: Move back to `hovertemplate` in Plotly charts
After extensive investigation by @jacky5124

### LINKED ISSUE #634: Write `hovertemplate` for Plotly plots to serve as backup
The `hovertemplate` field of Plotly plot objects, as mentioned by #307 , is the default option of any Plotly plot when more detail is needed to display whenever a data point on a plot is clicked or hovered. Even tough our custom hover card is getting better and better, like this update by #626 , there may still be situations that prove using the custom one is tricky, such as exporting the plots as HTML so that other websites can embed the same plots. Therefore, we should still create a default `hovertemplate` for each Plotly plot, and add a condition that controls whether the default option or the custom option is used.

jacky5124:
Fixes #634 , fixes #603 

For household output plots, the hover messages are simple such that they are in "if x then y" format, so `hovertemplate` for such plots is just a template.

But for policy reform plots, the hover messages are dynamic such that the entire messages can change depending on the x and/or the y value, so a simple template no longer works for such plots, and `customdata` is required to pre-compute the messages, which will be available in `hovertemplate` as a variable like x and y.

Added a prop `useHoverCard` to every plot component so that switching between Plotly's `hovertemplate` and the custom HoverCard is as easy as passing a boolean value to the component.

household

policy

### Summary

No summary available (Limit exceeded: required to process 64082 tokens, but only 50000 are allowed per call)

### Walkthrough
No walkthrough available (Limit exceeded: required to process 64082 tokens, but only 50000 are allowed per call)


jacky5124:
Apology for closing this pull request for a while. The entire code base was linted few days ago and it had conflict with the original commits in this PR, but I accidentally rebased this PR's branch instead of merging with master branch, which caused parallel changes on master branch being recognized as my changes and messed up things. Now the changes can be properly merged with master branch.

nikhilwoodruff (approved):
Amazing- apologies for the lateness and a belated thanks!

PR #703, opened 2023-08-20
Update US Default object for the calculate route request

### LINKED ISSUE #702: Issue with PolicyEngine status page for US Calculate
The US `calculate` route was showing an error, which did not match the results shown from Better Stack.

image

samau3:
Fixes #702 

The status information for the `calculate` endpoints is determined by making a request to the endpoints with basic household data as an object with key value pairs. So if the endpoints have changed in regards to what keys and/or values it can accept, then this status page may incorrectly show an `Error` for the endpoint route. Therefore, the  `DEFAULT_US_HOUSEHOLD` object needed to be updated to better match the expected data structure for the API.

### DIFF
--- a/src/pages/statusPageDefaults.js
+++ b/src/pages/statusPageDefaults.js
@@ -29,335 +29,11 @@ export const STATUS_TEXT_COLORS = {
 };
 
 const DEFAULT_US_HOUSEHOLD = {
-  families: {
-    "your family": {
-      family_id: {
-        2023: 0,
-      },
-      family_weight: {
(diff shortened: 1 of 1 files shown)


MaxGhenis:
Will we need to update this each time we add/remove a variable from `policyengine-us`?

anth-volk:
This is now engineered more flexibly so that the default household on the front end starts as an empty object, then variables are added with their default values to prevent breaking when variables are renamed in the country packages. Closing.

PR #826, opened 2023-10-31
Fixes #771: Hide advisory board from About page

### LINKED ISSUE #771: Hide advisory board from About page
Pending an announcement including some new members

JiginJayaprakash:
copilot:all
Fixes #771
added a div with display none to hide all the advisory


nikhilwoodruff (approved):
Thanks @JiginJayaprakash!

PR #834, opened 2023-11-03
Update API screenshot on homepage.

### LINKED ISSUE #215: Use fixed width font for code demo on homepage
image

abhcs:
Fix #215 by updating the API screenshot image file rendered on the homepage.

### DIFF
--- a/src/redesign/components/APIDocumentationPage.jsx
+++ b/src/redesign/components/APIDocumentationPage.jsx
@@ -365,6 +365,7 @@ export default function APIDocumentationPage({ metadata }) {
         <p>Try out the API in this interactive demo.</p>
         <iframe
           src={`https://policyengine-policyengine-api-demo-app-xy5rgn.streamlit.app/~/+/?mode=${countryId}`}
+          // the demo is in the policyengine-api-demo repository in PolicyEngine
           title="PolicyEngine API demo"
           height="500px"
           width={"100%"}


anth-volk (changes requested):
Thanks for your contribution! Unfortunately, I fear this does the opposite of what #215 is looking for - what we're hoping to do is change the font of the code output screen on that page so that it uses a monospace font like Courier instead of the Arial or Helvetica it's currently using. The image on the homepage is actually correct; the item with the font issue has moved from the pre-redesign homepage to a new API page, which may be part of the confusion.

Would you be willing to look into why the API renders this code with the Arial and change it to monospace?

abhcs:
Initial investigation suggests that the code responsible for the font is in the [policyengine-api-demo](https://github.com/PolicyEngine/policyengine-api-demo) repository:

https://github.com/PolicyEngine/policyengine-api-demo/blob/353be7075283c58e2a61e0d6d2d1b6e70a4b8e28/app.py#L30

If streamlit is reading that repository, then changing the font there should fix the problem.

abhcs:
The issue is fixed [here](https://github.com/PolicyEngine/policyengine-api-demo/pull/1). In this PR, I have reverted previous commits and added a comment to help others find the policyengine-api-demo repository.

anth-volk:
Thanks so much. I don't have write access over there, so I linked Nikhil, hopefully he can address at a later point, or if I can be added, I'll check it out a bit later.

anth-volk:
I actually do have write access now. As I mentioned over there in the PR, would you mind uploading an image of the sample API code block with the formatting changes?

anth-volk:
Thanks for this, @abhcs. I approved the fix over at `policyengine-api-demo`, and I'll approve this, as it's nice to have the added comment.

anth-volk (approved):

Issue #17, opened 2022-11-29
Adjust y axis bounds on parameters based on entry


MaxGhenis:
Example: [this](https://beta.policyengine.org/uk/policy?household=774&focus=gov.contrib.ubi_center.basic_income.amount.flat&reform=160&region=uk&timePeriod=2022&baseline=1) should adjust to at least £100 when entering that value:
<img width="873" alt="image" src="https://user-images.githubusercontent.com/6076111/208285349-f2e85527-66b1-4af2-b80e-0a3b98da8ec0.png">

PR #839, opened 2023-11-11
Reflect policy parameter changes in right hand pane and middle chart

### LINKED ISSUE #53: Parameter remains in cart after resetting to original value
[For example](https://beta.policyengine.org/us/policy?reform=90&focus=gov.ssa.ssi.income.exclusions.earned_share&region=us&timePeriod=2022&baseline=2)
image

abhcs:
## Description

Fixes #53 and #17. The issues are about accurately reflecting changes made through the policy editor inputs (dates, value) in the right sidebar and parameter-over-time chart.

## Strategy for fix

The policy right sidebar (#53) and the parameter-over-time chart (#17) display the information in policy.reform.data, directly or via the getReformedParameter() function. I have modified getReformedParameter() so that the end date is also inserted into the reformed parameter, and I have used getReformedParameter() to construct new policy objects (see newReforms() in ParameterEditor.jsx). These changes ensure that the correct information is available to the right sidebar and the chart.

I have modified the input field component so that it does not store unnecessary state and calls the onChange function with the placeholder when the value is deleted from the field. This change was needed for #53.

I have added new logic for generating axis ranges for dates, and used it to update the parameter-over-time chart. This change helps with #17.

## Tests

I have checked the behavior of the sidebar and the chart for policy inputs of different types (bool, %, numeric).

copilot:all

### DIFF
--- a/src/api/parameters.js
+++ b/src/api/parameters.js
@@ -101,19 +101,38 @@ export function getReformedParameter(parameter, reforms) {
   if (!parameterValues) {
     return null;
   }
-  const parameterValuesInOrder = Object.keys(parameterValues).sort();
+  const sortedKeys = Object.keys(parameterValues).sort();
   const reform = reforms[parameter.parameter];
   if (reform) {
     for (const [timePeriod, value] of Object.entries(reform)) {
       const [startDate, endDate] = timePeriod.split(".");
-      // Delete all values in the time period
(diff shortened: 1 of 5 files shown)


nikhilwoodruff:
Thanks for this @abhcs! Just tested it out locally and it's a big improvement.
image

nikhilwoodruff (approved):

PR #858, opened 2023-11-22
refactor Open Collective div into link

### LINKED ISSUE #776: Link to Open Collective is not a true link, disabling right-click options
For accessibility and convenience reasons, it'd be good to re-engineer this so that it's a true link

nataliia-reshetnikova:
Closes #776:

Changes:
- refactor button "Donate on Open Collective" into <a> link element so the right click works on it
- create constants for emails and links on the Donate page that are repeating
- refactor repeating text into a separate function

Results of npm test:
1 fails but not related to changes in this PR:
"Test updateHousehold function › Ensure function removes variables properly from US household"

Suggestion for improvement:
- "hello@policyengine.org" is repeating multiple times on the website and in the codebase, so it might be a good idea to make it a constant for future maintanance.
- if there are other elements that would benefit from refactoring from clickable div into a link, could create a component for this purpose

### DIFF
--- a/src/redesign/components/Donate.jsx
+++ b/src/redesign/components/Donate.jsx
@@ -4,15 +4,52 @@ import Section from "./Section";
 import style from "../style";
 import PageHeader from "./PageHeader";
 import useDisplayCategory from "./useDisplayCategory";
-import ActionButton from "./ActionButton";
 
 export default function Donate() {
   const displayCategory = useDisplayCategory();
   const mobile = displayCategory === "mobile";
   const desktop = displayCategory === "top";
+  const pslFoundUrl = "https://psl-foundation.org";
+  const openCollectiveUrl = "https://opencollective.com/psl-foundation";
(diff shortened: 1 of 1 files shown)


nikhilwoodruff:
Thanks @nataliia-reshetnikova for neatening up the codebase here!

nikhilwoodruff (approved):

MaxGhenis:
Thank you @nataliia-reshetnikova and sorry we let this get stale. I had to make some content changes today for Giving Tuesday and now we have significant merge conflicts. But I like how you approached this, would you like to revise it, or prefer to start fresh?

Also cc @nung22 who took a different stab in https://github.com/PolicyEngine/policyengine-app/pull/909

Could you two take a look at these and suggest a strategy? Thanks!

nataliia-reshetnikova:
Thank you for the feedback, @MaxGhenis , @nikhilwoodruff.  I think since there were significant changes on the page, it is better just to close this one.

PR #863, opened 2023-11-22
added an edge case to pass one of the unit tests

avelynhc:
fixes #852
- added an edge case in `src/pages/HouseholdPage.jsx` to fix the CI problem

PR #867, opened 2023-11-23
Fix useNavigate() runtime issue, thereby re-enabling cookie banner and policies modelled popup

anth-volk:
This commit fixes an error with the deployment of `useNavigate()`, re-enabling all interstitials. It also adds the cookie consent component back into the app, alters some styling on the old component, and adds tests. Finally, it re-enables the `PoliciesModelledPopup` and adds a test for it, as well.

Note: This also disables a test that currently always fails, as part of issue #836, until further planning can make the test more robust to back-end variable changes.

Fixes #762, #775, #866, and #817. If you'd prefer these broken into multiple PRs, happy to do so; I've lumped them together because each depends upon the other in the following order: solve useNavigate() -> refactor buttons -> re-style cookie banner in accordance with refactored buttons -> solve #775 using refactored buttons

### Summary
🧪🎨🍪

This pull request refactors the navigation and button components, adds new components for cookie consent and arrow buttons, adds new test files for buttons and dialogs, and fixes a test and a console.log statement. The overall purpose is to improve the user interface, the code quality, and the test coverage of the app.

> _We're refactoring the buttons, we're changing the props and styles_
> _We're testing the components, we're using the latest tools_
> _We're navigating the pages, we're linking to the resources_
> _We're coding with the fire, we're the masters of the `Button`!_

### Walkthrough
*  Add new dependencies for private class fields and custom jest matchers ([link](https://github.com/PolicyEngine/policyengine-app/pull/867/files?diff=unified&w=0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R85-R86))
*  Refactor Button component to use HoverBox, accept different types, and simplify props and styles ([link](https://github.com/PolicyEngine/policyengine-app/pull/867/files?diff=unified&w=0#diff-f8cf5ea2a1ff50689cdc818275b7d5a073177d6f4db6caecfbbf3aaf356dba55L1-R87))
*  Create ArrowButton component for navigation with arrow icons and hover effects ([link](https://github.com/PolicyEngine/policyengine-app/pull/867/files?diff=unified&w=0#diff-665f5abd5c44141f41bf7aabfec932aaddbcc1000ba710a13af111e74607d19fL1-R23))
[...]

Issue #762, opened 2023-10-21
useNavigate() error when testing other changes


anth-volk:
When testing another change that's both minor and not related to useNavigate(), I receive a useNavgiate() runtime error reading `useNavigate() may be used only in the context of a <Router> component.` The page is still accessible, but requires me to minimize a runtime error pop-up. Granted, this may be because of the changes I made while testing, but the similarity between it and an error Nikhil had received previously made me think it prudent to open this.

anth-volk:
Update: I now receive this error merely when running the `master` branch locally. This may still be an underlying issue.

anth-volk:
It appears that the reason behind this is as follows:

1. The `ActionButton` component, a new control component within the `redesign` folder, takes an optional prop used for navigation. It imports `useNavigate()` from `react-router-dom`, then instantiates an instance, but does not utilize it if the optional prop isn't passed.
2. The `Button` component builds on top of `ActionButton` but does not allow for the passage of link props. Even so, because `ActionButton` does, it builds on top of an active instance of `useNavigate()`.
3. The `PoliciesModelledPopup` (PMP) component, an interstitial message indicating what PolicyEngine does and does not model for each country context, utilizes a `Button` component to allow the user to minimize it. However, the PMP component does not exist within the `BrowserRouter` nested routes list, as it is not actually a page.

What seems to happen is that when the interstitial attempts to open, the `ActionButton` component attempts to instantiate an instance of `useNavigate()`, but is unable to do so, as PMP is not within the context of a `BrowserRouter` instance. The interstitial doesn't display while the runtime error window does, but because everything else on the site works, it's possible to still use the site as normal. I also wonder if maybe when `npm run build` is invoked, the build perhaps ignores the interstitial because of the runtime error, making it possible to deploy the site without visible errors (other than the interstitial not displaying), but making it impossible to debug without running into the runtime error.

PR #871, opened 2023-11-23
add a single command for handling both code linting and formatting.

### LINKED ISSUE #865: Add a command to handle both linting and formatting with Prettier
When running `npm run lint`, it prompts me to execute Prettier separately for formatting. It would be convenient if `npm run lint` could handle both linting and formatting without the need for an additional Prettier command.

Screenshot 2023-11-22 at 6 11 27 PM

avelynhc:
fixes #865
- `npm run lint` now can handle both linting and formatting without the need for an additional Prettier command.
- includes previous [pr-863](https://github.com/PolicyEngine/policyengine-app/pull/863)


anth-volk (commented):
I would want Max's thinking on the changes to HouseholdPage before feeling comfortable "signing off," so to speak. The reason `fips` fails is because it was a variable that every household had and its default value was 6. The variable was removed around Oct. 31, but because it had a truthy value, the code we used to update households should prompt every household created prior to Oct. 31-ish to re-create their household. It doesn't at the moment because issue #762 breaks the interstitial that's supposed to appear, making everything look okay. #762 should be resolved with PR #867. The front-end test fails at the moment because it pulls a real household for testing purposes, and since every US household had `fips`, it fails.

So here's what I'm thinking. We could merge this code, which will prevent household invalidation across the board and will make the test pass, but is a bit of an ugly workaround because we'll have to do it any time a variable like `fips` is deleted in the future. We could re-engineer the household object the test uses to make it hardier, but that will still invalidate all households. And/or, moving forward, we could have a discussion about how better to treat deleted truthy variables in this particular component.

Curious to hear @MaxGhenis's thoughts on this.

MaxGhenis:
Also would suggest scoping PRs to more specific issues, e.g. as we're discussing the lint/format change in #902

anth-volk:
@MaxGhenis Fair points on both

anth-volk (commented):
This was already solved and implemented via #902.

Issue #760, opened 2023-10-21
Organization logos are hard to read on US site


anth-volk:
The logos included in the "Trusted Across the US" bar are quite a bit smaller on the US site than the UK site; perhaps they're each longer on the US version? Additionally, it appears the width of the bar is fixed to only be about 50% of the entire page at desktop breakpoints. It may be better to widen that, especially if we add more organizations to it.

dotslashbit:
@anth-volk Is this still open? Asking as the current PR hasn't been updated since 3 weeks

anth-volk:
I'd say that at this point, yeah, it is. Thanks for looking to take this on, @dotslashbit! I'll get around to your other PRs either late today or early next week, as I have to finish something else first.

PR #872, opened 2023-11-23
Increase logo size and adjust container width

pkvijay65:
**Title:**
Enhance Logo Size and Layout in "Trusted Across the US" Section (#760)

Description:
Refined the sizing and distribution of logos within the "Trusted Across the US" section to address concerns about their scale compared to the UK site.

**Changes:**

- Enlarged logos for enhanced visibility and uniformity with the UK site.
- Applied space-around in the flex container for better spacing on wider screens.
- Shifted to a max-width approach for dynamic logo size adjustment.

**Visuals:**
**Before Change for US Site:**
image

**After Change for US Site:**
image

**Before Change for UK Site:**
image

**After Change for UK Site:**
image

**Discussion Point:**
While addressing the issue of the smaller logos on the US site, I also observed that the logos appear smaller on the UK site too. So, I have done resizing universaally and they look ok to me. Should our adjustments be universally applied across both sites for consistency, or should we consider separate sizing strategies for each site based on logo sizes for both the countries?

### DIFF
--- a/src/redesign/components/HomeUsedBy.jsx
+++ b/src/redesign/components/HomeUsedBy.jsx
@@ -41,10 +41,11 @@ export default function HomeUsedBy() {
           style={{
             display: "flex",
             justifyContent: {
-              mobile: "center",
-              tablet: "center",
-              desktop: "center",
+              mobile: "space-around",
+              tablet: "space-around",
+              desktop: "space-around",
             }[displayCategory],
+            flexWrap: "wrap",
             marginBottom: 30,
             marginTop: 20,
           }}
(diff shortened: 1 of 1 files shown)


anth-volk:
My personal view would be that it would be better if they were consistently the same size across sites. That said, it's always seemed to me that the logos on the US site look smaller than the ones on the UK. At present, I'm wondering if we size based purely on width, or if we treat each as a square and if maybe that causes the US ones to look smaller.

MaxGhenis:
I agree with consistency across countries although I don't think anyone except the dev team looks at multiple countries, so if this change makes each country homepage look better without creating significant gaps across countries it seems worth merging

anth-volk:
Hey @pkvijay65, apologies for the delay in re-addressing this. Would you be willing to rebase off main? If there are any merge conflicts, I'm happy to handle those

anth-volk (changes requested):
Hey @pkvijay65, apologies for the delay in re-addressing this. Would you be willing to rebase off main? If there are any merge conflicts, I'm happy to handle those

anth-volk:
Closing as stale in favor of contributions from a different contributor.

PR #909, opened 2023-11-28
Apply Hover Effect to "Donate on Open Collective" button 

nung22:
### Summary
🐛🖥️🎁

Fix a bug in the donate button component by correcting the value of a variable. This ensures that the button displays correctly on different screen sizes.

> _`desktop` variable_
> _changed to fix donate button_
> _a winter blessing_

### Walkthrough
* Fix a bug that prevented the desktop version of the donate button from rendering properly by changing the value of the `desktop` variable to match the possible values of the `displayCategory` hook ([link](https://github.com/PolicyEngine/policyengine-app/pull/909/files?diff=unified&w=0#diff-748bcc9af789f94cdf4dc4b4ecaa73de6709a51dcff57cc2b44ef5af55b59236L12-R12))

### Hover effect now working

issue-815-gif1


anth-volk:
Hey @nung22, sorry for the delay in reviewing this, but there were some pretty significant changes to the codebase relevant to your PR. Would you be willing to close this PR, then restart/rebase from master? 

There's actually a broader issue in the underlying `Button` component at the moment where it's composed of one or two divs and where the hover effect cannot be applied on percentage-based widths due to the technical limitations of CSS. Additionally, there's a broader discussion about whether we should replace this hover effect with something instantaneous. At least in the meantime, would you be willing to address the first one by rewriting the Button component as a Button and working on a hover effect that works the same way, but can be applied to percentage-sized buttons? This is a lower-priority issue, but it would be nice to have some consistency.

anth-volk:
I'm going to close this for being stale for a few reasons:

1. This would require significant rebasing and merge-conflict fixing, as we've changed the buttons quite a bit and changed the link this button points to.
2. There are more buttons that don't function like this, and it's driven by inherent limitations in the Button class itself.
3. We're discussing swapping the buttons out for a standard library, anyway, so the time spent reworking these would be wasted.

Issue #928, opened 2023-12-02
Remove decimals from age input box


MaxGhenis:
example https://policyengine.org/us/household?focus=input.household.age

<img width="674" alt="image" src="https://github.com/PolicyEngine/policyengine-app/assets/6076111/d7787cfe-c674-4dec-806c-0524e2feae01">

abhcs:
@MaxGhenis I am not sure that age is an int. The backend reports to the front end that it is a float, and its float-ness is used in variables such as ccdf_age_group.

MaxGhenis:
You're right my mistake! Will check after updating -us

PR #941, opened 2023-12-07
Format currency input with commas in household input

### LINKED ISSUE #178: Format currency in real time, with commas
Currency inputs are currently floats that should not have commas (see #143). Similar to #177, let's format numbers in real time, e.g. as a user enters `2013` it shows `$2,013`, or if they enter `2013.3` it shows `$2,013.30`.

abhcs:
Fixed #178 by using the formatter property in ant-design InputNumber component. Renamed locale en to en-GB, the standard Unicode identifier for the English language in the UK region (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale for details). Improved the interface for locale-related functions with better names and documentation.

Tested in UK and US locales to verify that currencies are displayed properly in inputs and charts.
Screenshot 2023-12-07 at 11 41 45 AM

Screenshot 2023-12-07 at 11 44 44 AM

### Summary
🌐💷📊

This pull request improves the localization of currency values and plotly charts in the policy engine app. It introduces two new functions, `localeCode` and `currencyCode`, to get the appropriate codes for a given country. It also updates the `VariableEditor` component and the output charts to use these functions and the native `toLocaleString` method of numbers. It also renames and updates the UK English plotly locale module.

> _Oh, we're the jolly coders of the policy engine_
> _We format currencies and locales with ease_
> _We use `localeCode` and `currencyCode` to make it clean_
> _And we heave away on the pull request with a breeze_

### Walkthrough
*  Replace `currencyString` and `localeString` functions with `localeCode` and `currencyCode` functions in `src/api/language.js` to return locale and currency codes for a given country id ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-77a1d7924c7e86c45476b6d602f1891a692c24c0f5d85a55f3b7df008d6803b3L51-R59))
  - In `src/pages/household/input/VariableEditor.jsx`, update the `formatter` prop of the `InputNumber` component in the `HouseholdVariableEntityInput` function ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-0b96b3d13d96221f5cb68df970385b96da6e895d2437c2fd666a3d052401b991L2-R2), [link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-0b96b3d13d96221f5cb68df970385b96da6e895d2437c2fd666a3d052401b991L235-R239))
  - In `src/pages/policy/output/AverageImpactByDecile.jsx`, update the `text` prop of the `bar` trace in the `AverageImpactByDecilePlot` function ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-ec16d5261d93e698e4741e7df864ae2f82625d02047ca9697db51ae269ee05a2L4-R4), [link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-ec16d5261d93e698e4741e7df864ae2f82625d02047ca9697db51ae269ee05a2L36-R42))
  - In `src/pages/policy/output/AverageImpactByWealthDecile.jsx`, update the `text` prop of the `bar` trace in the `AverageImpactByWealthDecilePlot` function ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-f561e2313f626b35f5ba151c460ff2c4a2919ef88e0a5ba86fd61f60a0ac4b8aL4-R4), [link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-f561e2313f626b35f5ba151c460ff2c4a2919ef88e0a5ba86fd61f60a0ac4b8aL36-R42))
  - In `src/pages/policy/output/AverageImpactByDecile.jsx`, update the `locale` prop of the `Plot` component in the `AverageImpactByDecilePlot` function ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-ec16d5261d93e698e4741e7df864ae2f82625d02047ca9697db51ae269ee05a2L105-R111))
  - In `src/pages/policy/output/AverageImpactByWealthDecile.jsx`, update the `locale` prop of the `Plot` component in the `AverageImpactByWealthDecilePlot` function ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-f561e2313f626b35f5ba151c460ff2c4a2919ef88e0a5ba86fd61f60a0ac4b8aL105-R111))
  - In `src/pages/policy/output/BudgetaryImpact.jsx`, update the `locale` prop of the `Plot` component in the `BudgetaryImpactPlot` function ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-af88a6bb0672282e83959a0a8317e5259f5e148af72f4901f431073ca751e7b3L4-R4), [link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-af88a6bb0672282e83959a0a8317e5259f5e148af72f4901f431073ca751e7b3L159-R159))
  - In `src/pages/policy/output/DetailedBudgetaryImpact.jsx`, update the `locale` prop of the `Plot` component in the `DetailedBudgetaryImpactPlot` function ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-d50bb4f3433b375516906846f34ac9a1d86d9d361680f9d47cf9cf58f4a44cf5L4-R4), [link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-d50bb4f3433b375516906846f34ac9a1d86d9d361680f9d47cf9cf58f4a44cf5L132-R132))
* Rename `src/plotly_locales/locale-en.js` to `src/plotly_locales/locale-en-gb.js` and change the `name` property of the `loc_en` module to `en-GB` to reflect the correct locale code for the UK English plotly locale ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-453d87bff954264f9dbeea0da5ea8c8791a7236c303799ff647303f93ebb23d6L7-R7))
* Update the import and registration of the UK English plotly locale in `src/redesign/components/PolicyEngine.jsx` to use the renamed `loc_en_gb` module ([link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-e30e417fc83ec5700443c56bd7a183339e16b3094194b57c61fdbd596eb7d2d8L23-R23), [link](https://github.com/PolicyEngine/policyengine-app/pull/941/files?diff=unified&w=0#diff-e30e417fc83ec5700443c56bd7a183339e16b3094194b57c61fdbd596eb7d2d8L43-R43))

### DIFF
--- a/src/api/language.js
+++ b/src/api/language.js
@@ -48,14 +48,13 @@ export function cardinal(number) {
   return number + (suffixes[(rem - 20) % 10] || suffixes[rem] || suffixes[0]);
 }
 
-export function currencyString(number, metadata, maximumFractionDigits) {
-  return number.toLocaleString("en-US", {
-    style: "currency",
-    currency: metadata.countryId === "uk" ? "GBP" : "USD",
-    maximumFractionDigits: maximumFractionDigits,
-  });
+// returns the Unicode locale identifier for the country id in the metadata
+export function localeCode(countryId) {
+  return countryId === "uk" ? "en-GB" : "en-US";
 }
 
-export function localeString(metadata) {
-  return metadata.countryId === "uk" ? "en" : "en-US";
+// returns the ISO 4217 currency codes for the currency for the country id in
+// the metadata
+export function currencyCode(countryId) {
+  return countryId === "uk" ? "GBP" : "USD";
 }
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -1,5 +1,5 @@
 import { useSearchParams } from "react-router-dom";
-import { capitalize } from "../../../api/language";
+import { capitalize, localeCode } from "../../../api/language";
 import {
   currencyMap,
   getNewHouseholdId,
@@ -232,7 +232,11 @@ function HouseholdVariableEntityInput(props) {
           width: mobile ? 150 : 200,
         }}
         addonBefore={isCurrency ? currencyMap[variable.unit] : undefined}
-        min={isCurrency ? 0 : undefined}
+        formatter={
+          isCurrency
+            ? (value) => (+value).toLocaleString(localeCode(metadata.countryId))
+            : undefined
+        }
         precision={variable.valueType === "float" ? 2 : undefined}
         defaultValue={defaultValue}
         autoFocus={true}
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -223,6 +223,7 @@ function HouseholdVariableEntityInput(props) {
     }
   }
   const mobile = useMobile();
+  const precision = 2;
   let control;
   if (variable.valueType === "float" || variable.valueType === "int") {
     const isCurrency = Object.keys(currencyMap).includes(variable.unit);
@@ -231,15 +232,18 @@ function HouseholdVariableEntityInput(props) {
         style={{
           width: mobile ? 150 : 200,
         }}
-        addonBefore={isCurrency ? currencyMap[variable.unit] : undefined}
-        formatter={
-          isCurrency
-            ? (value) => (+value).toLocaleString(localeCode(metadata.countryId))
-            : undefined
-        }
-        precision={variable.valueType === "float" ? 2 : undefined}
+        {...(isCurrency
+          ? {
+              addonBefore: currencyMap[variable.unit],
+              formatter: (value) =>
+                (+value).toLocaleString(localeCode(metadata.countryId), {
+                  maximumFractionDigits: precision,
+                }),
+            }
+          : {})}
+        {...(variable.valueType === "float" ? { precision: precision } : {})}
         defaultValue={defaultValue}
-        autoFocus={true}
+        autoFocus
         onChange={submitValue}
       />
     );
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -238,13 +238,13 @@ function HouseholdVariableEntityInput(props) {
           : {})}
         {...(variable.valueType === "float"
           ? {
-              formatter: (value) =>
+              formatter: (value, {userTyping}) =>
                 (+value).toLocaleString(localeCode(metadata.countryId), {
-                  maximumFractionDigits: 16,
+                  minimumFractionDigits: !userTyping && 2,
+                  maximumFractionDigits: 2,
                 }),
             }
           : {})}
-        precision={2}
         defaultValue={defaultValue}
         autoFocus
         onChange={submitValue}
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -240,10 +240,11 @@ function HouseholdVariableEntityInput(props) {
           ? {
               formatter: (value) =>
                 (+value).toLocaleString(localeCode(metadata.countryId), {
-                  maximumFractionDigits: 2,
+                  maximumFractionDigits: 16,
                 }),
             }
           : {})}
+        precision={2}
         defaultValue={defaultValue}
         autoFocus
         onChange={submitValue}
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -223,7 +223,6 @@ function HouseholdVariableEntityInput(props) {
     }
   }
   const mobile = useMobile();
-  const precision = 2;
   let control;
   if (variable.valueType === "float" || variable.valueType === "int") {
     const isCurrency = Object.keys(currencyMap).includes(variable.unit);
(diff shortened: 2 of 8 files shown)


nikhilwoodruff:
Thanks @abhcs - how does this handle user input? Just to check that if a user sees commas and changes the value to a string including commas that this wouldn't break the app.

abhcs:
The formatter property changes the display string on each event. The user will be unable to type the string `1,234` in the normal way because at `1,2` the output will get reformatted to `12`.

| input | display |
| ----------- | ----------- |
| `1`      | `1`       |
| `1,`   | `1,`        |
| `1,2`   | `12`        |
| `1,23`   | `123`        |
| `1,234`   | `1,234`        |

Note that the current component also reformats the display string similarly but does not add commas. This is an unavoidable side-effect of "formatting in real time" (#178); see also the formatter example at https://ant.design/components/input-number.

If one were to directly paste the string `$123456` into the new component, then it would get reformatted to `123,456`. In the current component, it would get reformatted to `123456`.

anth-volk:
This looks like a great change. I tested locally and can see that your edits do handle the comma properly, as well as clipping out garbage values. That said, per the video I've included here, the edits don't properly format an input sub-dollar unit divisible by ten; for instance, if I input "$15,000.30", it shortens this to "$15,000.3". So, to proceed, I'm wondering if you'd be willing to do the following:

1. Investigate if there is an antd setting that always displays two decimal places if the input number contains a decimal point (at least after the user is no longer focused on the box)
2. Remove the default "0" that's displayed when nothing has been input, and instead have the box be empty, as it is currently on the live app
3. Rebase off of master

Thanks so much for your contributions to the app!

anth-volk:
https://github.com/PolicyEngine/policyengine-app/assets/14987227/19d2752a-e29d-47a0-9a39-bfa5ddad8383

Hopefully the video uploaded correctly in this comment

anth-volk (changes requested):
Apologies, I should've left my comments here, but please see my comments on the PR itself

abhcs:
> This looks like a great change. I tested locally and can see that your edits do handle the comma properly, as well as clipping out garbage values. That said, per the video I've included here, the edits don't properly format an input sub-dollar unit divisible by ten; for instance, if I input "$15,000.30", it shortens this to "$15,000.3". So, to proceed, I'm wondering if you'd be willing to do the following:
> 
> 1. Investigate if there is an antd setting that always displays two decimal places if the input number contains a decimal point (at least after the user is no longer focused on the box)

In the `OnChange` handler for `InputNumber`, we get a value of type `number`. The conversion of the input to this number is handled by antd; thus, the internal logic of `InputNumber` is responsible for the _good behavior_, e.g., removing elements like `$`. The formatter never sees the input `$15,000.30`, it only sees the number `15000.3` provided by antd. There are properties like `stringMode` in `InputNumber` that may be worth exploring but we would lose the abstraction of `InputNumber`.

The abstraction of InputNumber is an important one. Note in the image how the zeros after the comma are ignored when we just use `Input` in `ParameterEditor`:

Screenshot 2023-12-10 at 3 34 19 PM

> 2. Remove the default "0" that's displayed when nothing has been input, and instead have the box be empty, as it is currently on the live app

Since pull request #892, the default is displayed as `0.00`. The current request changes this to `0`.

> 3. Rebase off of master.
> 

Done.

> Thanks so much for your contributions to the app!
> 

You are welcome.

abhcs:
The new commits should fix #928. Now, I am using the formatter property for all floats. Despite age being a float, the trailing zeros get stripped.

anth-volk:
I've been digging into this to see if we can get it to display two decimal places only after the user clicks outside the input, as doing it before prevents the user from inputting a second decimal place without hitting backspace. I'll come back to it tomorrow.

abhcs:
Please check the latest commit. It may do what you are looking for; otherwise, we can revert it. It uses a strange combination of formatter and precision, which seems to work.

anth-volk (approved):
I've taken the liberty of making a minor edit: In Ant Design, the `precision` feel is overridden whenever a `formatter` is present, meaning your last commit unfortunately didn't have its intended impact. However, there's a pretty hidden `userTyping` param that allows for conditional formatting based on whether the user's actively engaging with the box, so I've edited it to format decimals only when the user isn't typing.

Thanks again for digging deep into antd's docs, @abhcs.

Edit: Just waiting to push until the checks finish

abhcs:
The docs say that `precision` is overridden by `formatter` but the solution was working! It was allowing the user to type an arbitrary number of fractional digits but the precision was getting applied on blur, i.e., not typing. I think that using `userTyping` is a better solution though. See PR #962.

anth-volk:
Very strange, perhaps your version was different from mine? When I installed and ran yours locally, the overriding did occur…not sure why.

abhcs:
Yeah, its best not to depend on undocumented quirks in any case. So `userTyping` is a better solution. Maybe it was browser-specific. I am using Safari.

anth-volk:
Ah, that could be it. We’ve had cases before where Safari will do one thing and all other browsers do another for some reason.

PR #955, opened 2023-12-12
Make state dropdown box searchable again

### LINKED ISSUE #954: Unable to type to search in the household calculator state dropdown box
Screen Shot 2023-12-11 at 6 09 23 PM

For this one here, there's no way to type and filter what state you want to choose. I wouldn't be surprised if this were also true for other dropdowns in the calculator interfaces.

abhcs:
Fix #954.

This regression was introduced in pull request #892. I have been trying to use ant-design components directly instead of the existing wrappers in the app such as `SearchOptions` because the ant-design components have better interfaces and documentation than the wrappers in the app. Usually, well-designed components provided by a UX framework do not need wrappers and should be used directly. I will be more careful with switching components in the future.

### Summary
🔎🏷️🚀

Added search and filter functionality to the variable selector in `VariableEditor.jsx`. This makes it easier to find and choose a variable from the policy engine.

> _`Select` component_
> _Search and filter by label_
> _Autumn leaves the rest_

### Walkthrough
*  Enable searching and filtering variables by label in the editor ([link](https://github.com/PolicyEngine/policyengine-app/pull/955/files?diff=unified&w=0#diff-0b96b3d13d96221f5cb68df970385b96da6e895d2437c2fd666a3d052401b991R254-R255))

### DIFF
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -251,6 +251,8 @@ function HouseholdVariableEntityInput(props) {
   } else if (variable.valueType === "Enum") {
     control = (
       <Select
+        showSearch
+        optionFilterProp="label"
         style={{ width: mobile ? 150 : 200 }}
         options={variable.possibleValues}
         defaultValue={defaultValue}


anth-volk (approved):
Perfect! Thanks for adding this. Can definitely understand the move toward using antd out-of-the-box; we'll need to clarify if we're going to use them exclusively moving forward.

PR #958, opened 2023-12-12
Refactor: Donate.jsx

amar9312:
Resolves issue [776](https://github.com/PolicyEngine/policyengine-app/issues/776)

Changes:
* Wrap `<LinkButton` inside `<Link>`
* Set `link` prop of `<LinkButton` to `""`
* Remove `noArrow` prop of `<LinkButton`

npm test:
`No tests found related to files changed since last commit.`

### DIFF
--- a/src/__tests__/Buttons.test.js
+++ b/src/__tests__/Buttons.test.js
@@ -30,21 +30,4 @@ describe("Test button types", () => {
     fireEvent.click(getByText("Test Button"));
     expect(mockUseNavigate.mock.calls[0][0]).toBe("/test");
   });
-  test("Test that LinkButton properly navigates to external link", () => {
-    window.open = jest.fn();
-
-    const testButton = (
-      <LinkButton text="Test Button" link="https://www.google.com" />
-    );
-
-    const { getByText } = render(testButton);
-
-    fireEvent.click(getByText("Test Button"));
-    expect(window.open).toHaveBeenCalledWith(
-      "https://www.google.com",
-      "_blank",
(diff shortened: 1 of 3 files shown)


anth-volk:
Will be getting to this tomorrow. I think we can extrapolate this change and simply implement it within LinkButton itself. Thanks for this, really clean fix.

amar9312:
> Will be getting to this tomorrow. I think we can extrapolate this change and simply implement it within LinkButton itself. Thanks for this, really clean fix.

I agree with that. The LinkButton should have this feature.

anth-volk:
Apologies on the delays, but will get to this early next week

anth-volk (approved):
I took the liberty of simply moving your changes up to the LinkButton component itself, but the logic behind it remains yours. Thanks so much for your work @amar9312.

anth-volk:
This fails a test that was written against the old implementation. If you'd like to change it, please feel free, but I'm also happy to do so, likely tomorrow

anth-volk (changes requested):
Need to change LinkButton's test

anth-volk (approved):
Thanks again for the contributions, @amar9312. Now that this is factored out into the overall component, we've added a bit more accessibility to the app overall.

MaxGhenis:
Thanks @amar9312! Heads up if you type `Resolves #776` or `Fixes #776` in the first comment, it will link the issue so that merging the PR closes the issue. The way you did it does not link it, so I closed the issue manually.

PR #962, opened 2023-12-14
Remove decimals from age input box

### LINKED ISSUE #928: Remove decimals from age input box

abhcs:
Fix #928 by removing trailing zeros if the value is integral.

The change also allows the user to type a value with up to `16` fractional digits and truncates the value only on blur (`16` is chosen arbitrarily, of course). This is the more natural behavior because the inability to see the third fractional digit while it changes the second visible fractional digit is confusing. The simple ant-design `InputNumber` component with `precision = 2` also allows the user to enter many fractional digits and truncates the value on blur.

### Summary
🐛🎨🚸

Fixed a formatting bug and improved user typing for float variables in the `VariableEditor` component. Modified the formatter function in `src/pages/household/input/VariableEditor.jsx`.

> _`float` formatter_
> _bug fixed with more logic_
> _spring cleaning the code_

### Walkthrough
*  Fixed a bug in the formatter function for float variables to handle integer values and user typing better ([link](https://github.com/PolicyEngine/policyengine-app/pull/962/files?diff=unified&w=0#diff-0b96b3d13d96221f5cb68df970385b96da6e895d2437c2fd666a3d052401b991L241-R248)). This change improved the user interface and usability of the `VariableEditor` component, which allows users to edit the values of policy parameters and household characteristics in the `household` page.

### DIFF
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -240,11 +240,14 @@ function HouseholdVariableEntityInput(props) {
           : {})}
         {...(variable.valueType === "float"
           ? {
-              formatter: (value, { userTyping }) =>
-                (+value).toLocaleString(localeCode(metadata.countryId), {
-                  minimumFractionDigits: !userTyping && 2,
-                  maximumFractionDigits: 2,
-                }),
+              formatter: (value, { userTyping }) => {
+                const n = +value;
+                const isInteger = Number.isInteger(n);
+                return n.toLocaleString(localeCode(metadata.countryId), {
+                  minimumFractionDigits: userTyping || isInteger ? 0 : 2,
+                  maximumFractionDigits: userTyping ? 16 : 2,
+                });
+              },
             }
           : {})}
         defaultValue={defaultValue}


anth-volk:
I'm going to hold off on reviewing this until reviewing #975, as merging this may create merge conflicts, since the prod repo now contains a different input component

anth-volk:
@abhcs Now that #975 has been incorporated, would you mind rebasing off master before I review this?

anth-volk (commented):
Thanks for your changes on this. I get where you're coming from on this, but it looks like `isInteger` changes whenever the user changes the input to a floating-point value, making the age box merely display a float rounded to two variables. The critical issue here, though, is that `age`'s `value_type` in the metadata is described as `float`', not `int` for some reason. I'm going to open an issue, since it'd be ideal to actually apply universal rules for all `int` types instead of trying to determine if something's a float or int on the front end.

abhcs:
There is an issue open for that: https://github.com/PolicyEngine/policyengine-us/issues/3357. However, I think that age is not an int, because there are policies in the back end that require age to be float. I have mentioned one age variable of this type in #962: ccdf_age_group.

abhcs:
This PR fixes #962 in the following sense: the default value, if integral, will not contain decimals.

This PR also addresses another issue: the user can now freely type `12.99999`, and the value is rounded on enter or blur. On the live website trying to type `12.9999...` will keep incrementing the value in the box due to rounding.

anth-volk:
You're right, `age` is, in fact, required to be of type `float`, thanks for pointing that out. Yeah, this might just be the best workaround possible without altering `age` in the country packages to be of type `int`, then requiring variables in the packages ` to cast `age` from `int` to `float` (though that'd be my preference).

anth-volk (approved):
This is perhaps the best workaround to our own back-end limitations, and I do appreciate the additional effort to remove decimal processing that could confuse users. Thanks @abhcs.

PR #971, opened 2023-12-16
Disable microsim output temporarily

anth-volk:
An unfortunate temporary PR that disables all microsim output temporarily until we can diagnose an underlying issue. This PR returns the error page for both policy and household output. Temporarily fixes (?) policyengine-us [#3471](https://github.com/PolicyEngine/policyengine-us/issues/3471). The code commenting was necessary due to the linter.

copilot:all


MaxGhenis (approved):

MaxGhenis:
This also disables household activity (as intended)

Issue #967, opened 2023-12-14
The parameter input box does not parse strings with commas or currency symbols


abhcs:
Try changing the parameter in

https://policyengine.org/us/policy?focus=gov.contrib.ubi_center.basic_income.amount.person.flat&reform=40063&region=us&timePeriod=2023&baseline=2

to `1,100`. The string is parsed by the app as `1`. Now try `$11`. This is parsed as `0`.

Suggestion: We should use `InputNumber` here, like we do in `VariableEditor` for households.

anth-volk:
We should probably wait to work on this until solving #1001.

PR #975, opened 2023-12-16
Use a less reactive version of `InputNumber` in variable editor

### LINKED ISSUE #946: Restore the `InputNumber` component in `VariableEditor` and create new households less frequently when the user is editing a household
This issue is the household counterpart to issue #691. It is less severe than #691 since duplicates are not created, but more severe since we use more interactive input components like InputNumber.

## Steps to reproduce:

1. Create a household with income `x`.
2. Note the household id in the URL as `i`.
3. Edit the income using the up arrow key till it reaches `y`.
5. Note the household id in the URL as `j`.

The number of new households is `j-i` (assuming no one else is using the app). Observe that `j-i` is close to `y-x` in most cases.

## Expected output

We should see `j = i`, i.e., the household id should not change while the user is editing the current household. A new household should be created only when the user explicitly instructs the app to do so.

When the user asks the app to create a new household, the app should:
1. Show them the current household id asking them to note it down for future reference.
2. Ask them to confirm if they want to create a new household, warning that they may lose access to their previous work.
2. Allow them to back to editing the current household.

abhcs:
Fix #946 by triggering the creation of new households only on blur and on enter.

### DIFF
--- a/src/controls/StableInputNumber.jsx
+++ b/src/controls/StableInputNumber.jsx
@@ -1,0 +1,21 @@
+import { InputNumber } from "antd";
+import { useState } from "react";
+
+// The regular antd InputNumber is highly reactive: onChange is called whenever
+// the value is changed and on blur. In some situations (e.g., the
+// VariableEditor), we want to react to the new value only in response to
+// onPressEnter and onBlur events.
+export default function StableInputNumber(props) {
+  const { defaultValue, onChange, ...others } = props;
+  const [value, setValue] = useState(defaultValue);
+  const onPressEnter = () => onChange(value);
+  return (
+    <InputNumber
+      defaultValue={defaultValue}
+      onChange={setValue}
+      onPressEnter={onPressEnter}
+      onBlur={onPressEnter}
+      {...others}
+    />
+  );
+}
--- a/src/controls/StableInputNumber.jsx
+++ b/src/controls/StableInputNumber.jsx
@@ -1,20 +1,49 @@
 import { InputNumber } from "antd";
 import { useState } from "react";
 
-// The regular antd InputNumber is highly reactive: onChange is called whenever
-// the value is changed and on blur. In some situations (e.g., the
-// VariableEditor), we want to react to the new value only in response to
-// onPressEnter and onBlur events.
+/**
+ *
+ * The regular antd InputNumber is highly reactive: onChange is called whenever
+ * the value is changed through typing, scrolling, arrow keys, blur, etc. In
+ * some situations (e.g., in the VariableEditor), we want to react to the new
+ * value only in response to onPressEnter and onBlur events. This component is a
+ * thin wrapper over Ant Design's InputNumber component. Therefore, most
+ * properties for InputNumber are valid for StableInputNumber as well. There are
+ * a few differences:
+ *   1. The onPressEnter and onBlur callback functions have access to the
+ *      current value of the component.
+ *   2. The onChange handler is disabled.
+ *
+ * In the following toy example, a StableInputNumber component is created with
+ * initial value 10, and the current values in the field are logged to the
+ * console in response to the onPressEnter and onBlur events. The first argument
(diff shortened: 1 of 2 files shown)


anth-volk (changes requested):
Hey @abhcs, thanks for another great contribution. 

As I was reviewing this, it dawned on me that we currently use `InputField` in about five other spots in the application. More generally, there's been a desire to switch to a component library like `antd` more broadly to ensure app homogeneity and minimize bugs.

So here's what I'm thinking: would you be willing to change `StableInputNumber` into a component with a name akin to `FormattedInputNumber` or `FormattedNumericInput`. Ideally, this component would render the `antd` `InputNumber`, but providing the custom formatting as a default and provide a prop for formatting overrides (though that could be optional). I also envision props for `onChange` and something to designate `onSubmit` plus the clicking off action, something like `onSubmitLite`, but that's a terrible name for it. This way we could disambiguate some of the confusion currently in place in the app, where `onChange` in both the old and new code is not actually a change handler, but instead a hybrid of `submit` and `blur`.

This could then be inserted in place of `InputField` in other spots within the app, as I think we're going to want to maintain the behavior of this input app-wide. Feel free to let me know what you think: if this is too significant a feature, we can also scale it back. Thanks for all your effort, it's much appreciated.

abhcs:
From a search on the codebase, I could find uses of `InputField` only in `VariableEditor`, `ParameterEditor`, and `PolicyRightSidebar`. The use in `PolicyRightSidebar` does not seem like a right fit for `InputNumber`, and this PR already affects `VariableEditor`. That leaves `ParameterEditor` for which we have issue #967.

If I understand correctly, you are arguing for a higher-order component (HOC) called `FormattedInputNumber` that moves the logic for currency and formatting inside it. The `StableInputNumber` is also an HOC but has only one design function, making the ant-design component less reactive. The component proposed by you has more functionality packed inside it. This has pros and cons.

- Pros: `FormattedInputNumber` can be reused from `VariableEditor` and `ParameterEditor`. The design will satisfy the DRY ("Don't Repeat Yourself").
- Cons: An HOC that does too much is an anti-pattern. It's harder to read and maintain. Note how small `StableInputNumber` is, especially compared to `InputField`.

Given that `FormattedInputNumber` is expected to be used from 2-3 locations, is a HOC like `FormattedInputNumber` justified at the moment? Irrespective of how you answer this question, I think that we can address this design problem in #967.

anth-volk:
You make a fair point. My biggest quibble with the `StableInputNumber` implementation is that it doesn't address a bigger issue within the existing code base, i.e., that the `onChange` param being passed is not actually handling change events, but rather an abnormal mix of `blur` and `submit`.

I can see your logic behind not wanting to build out an entire HOC, especially since the third `InputField` in use is not actually of type `Number`. Would you instead be willing to either separately define an `onBlur` and `onSubmit` and pass the change handler to those (and maintain the built-in change handling in `StableInputNumber`)? In any future redesign of the calculator component, we can then build upon `StableInputNumber` to convert it into a more universal numeric input component if necessary, or we can leave it as it is.

MaxGhenis:
Meta suggestion but if this doesn't fix #946 as currently written, could you amend #946 to cover what this does, or link this PR to a new issue that this will fully resolve? Currently it's linked and will close #946 if we merge it.

anth-volk:
Thanks, this does look good @abhcs. One thing I will note is that it seems to slow loading times, but even when the user inputs a large variety of things very rapidly and switches to an output chart, the last value will populate correctly.

anth-volk (approved):

abhcs:
Please open an issue for the slow load time. I am using an anti-pattern here too, now that I think about it. The value is duplicated at two levels: inside `InputNumber` and in the wrapper. Thus, changing a value causes two renders instead of one. The reason for using the anti-pattern was because there is no straightforward way of accessing the value inside `InputNumber`.

anth-volk:
Ah, that does make sense. I'll open a new issue.

PR #977, opened 2023-12-18
Restrict the y-axis of the Marginal Tax Rates chart to `[-200%, 200%]`.

### LINKED ISSUE #66: Restrict MTR y axis at 100%

abhcs:
Fix #66 by filtering function values that exceed `200%` or go under `-200%`.

Tested against [this household](https://policyengine.org/us/household?focus=householdOutput.mtr&household=38027).

View when current income is `0`:

Screenshot 2023-12-18 at 10 25 23 AM

View when current income is `14,000`:

Screenshot 2023-12-18 at 10 26 52 AM

Thus, when the value corresponding to the current income does not lie in `[-200%, 200%]` the range on the y-axis is extended to include the value.

### DIFF
--- a/src/pages/household/output/MarginalTaxRates.jsx
+++ b/src/pages/household/output/MarginalTaxRates.jsx
@@ -39,43 +39,6 @@ export default function MarginalTaxRates(props) {
   const [showDelta, setShowDelta] = useState(false);
   const mobile = useMobile();
 
-  // We don't want to show "large" values in this chart. See
-  // https://github.com/PolicyEngine/policyengine-app/issues/66.
-
-  const lb = -2;
-  const ub = 2;
-
-  const clipBefore = (array) => {
-    if (array !== undefined && array.length) {
-      let i = 0,
-        min = Number.POSITIVE_INFINITY,
-        max = Number.NEGATIVE_INFINITY;
-      while (i < array.length && min > lb && max < ub) {
-        const element = array[i];
-        if (element < min) {
-          min = element;
-        }
-        if (element > max) {
-          max = element;
-        }
-        i++;
-      }
-      return [Math.max(lb, min), Math.min(ub, max)];
-    }
-    return [-1, 1];
-  };
-
-  const clipAfter = (format, currentValues) => {
-    const clip = [
-      Math.min(...currentValues, lb),
-      Math.max(...currentValues, ub),
-    ];
-    if (format !== undefined && Object.keys(format).includes("range")) {
-      const r = format.range;
-      format.range = [Math.max(clip[0], r[0]), Math.min(clip[1], r[1])];
-    }
-  };
-
   let title;
 
   const currentEarnings = getValueFromHousehold(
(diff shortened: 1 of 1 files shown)


abhcs:
There are two peaks above `200` but we ignore them, so the range of y-values is `[-90, 70]`. We add a padding of `0.1 x (70 - (-90)) = 16'. The padded range of `[-106, 86]' is displayed. The mentioned calculations are approximations.

anth-volk:
Gotcha. @MaxGhenis do you feel the 100% axis should be displayed if there are any values above it? I would personally lean toward it.

abhcs:
Just thinking about it a bit, if most of the values are within say `[0,10]` and there is one outlier of `700`, then the current design would keep the axis range around `[0,10]` with some padding, which is desirable, I think. In such a case, having a forced range of `[0,100]` would shrink the features.

The MTR function displayed in the chart is multidimensional with a vast domain. Still, we are making decisions about how to display the function based on a few samples from the domain, i.e., viewing the website for a particular household. This is a bit dangerous and the magic constants like `200` and `100` may not work for other households.

A more robust solution would be to find the outliers in the back end without relying on constants (I have code for this by using a Python lib for 1-D clustering). However, this would require changes to the API. Therefore, I have gone ahead with the current PR.

All that being said, I am happy to add the `100` constant too.

MaxGhenis:
Hi sorry I'm getting to this late. Can we instead set the y axis limits, rather than the values? So take the bounds of the data, add e.g. 10% on either side, and then clip those bounds at +/-200% for setting y axis limits, but leave the values as they are.

anth-volk (changes requested):
Per Max:

Can we instead set the y axis limits, rather than the values? So take the bounds of the data, add e.g. 10% on either side, and then clip those bounds at +/-200% for setting y axis limits, but leave the values as they are.

Thanks so much for your work on this, @abhcs.

abhcs:
We *are* changing axis limits and not values in this PR. In the two examples posted in the initial comment, do you expect to see something else? Please clarify.

anth-volk:
He had been thinking a modification to the code that, instead of merely filtering values outside of the bounds, would determine if values exist above and/or below, and then set the upper and lower bounds at +/- 2 accordingly, as merely filtering dynamically sets the chart range based upon the remaining values

abhcs:
Screenshots after the latest changes for the same household with different incomes:

For income = `0`:


For income = `14,000`:
14k

For income = `55,000`:
55k

MaxGhenis:
can we trim the 735% example too?

abhcs:
We can but the current behavior is by design! If we were to trim the 735% case, then 
- the dark blue circle representing the current MTR would not be visible on the graph, and
- double-clicking the current MTR legend would just yield an empty graph.

I have kept the current values visible in the chart to avoid this behavior. An alternative could be to not display the current value and the legend and still trim! Please let me know if you want this.

MaxGhenis:
Ok sounds good let's try this and we can come back to it if it's confusing

anth-volk (approved):
This looks great and works well on my side, @abhcs. Thanks so much for your contributions!

PR #979, opened 2023-12-18
Extend the cliffs in the Earnings Variation chart to meet its horizontal borders

### LINKED ISSUE #978: Extend the cliffs in the Earnings Variation chart to meet its horizontal borders
In this [example](https://policyengine.org/us/household?focus=householdOutput.earnings&household=38027), the cliffs get truncated at `0` and `150,000`:

Screenshot 2023-12-18 at 10 43 49 AM

abhcs:
Fix #978 by passing the y-axis range to the `getCliffs` function.

Tested against [this household](https://policyengine.org/us/household?focus=householdOutput.earnings&household=38027). Now, the chart looks like this:

Screenshot 2023-12-18 at 11 17 55 AM

### DIFF
--- a/src/api/variables.js
+++ b/src/api/variables.js
@@ -228,7 +228,7 @@ export function getPlotlyAxisFormat(unit, values, precisionOverride) {
   const paddedRange = () => {
     const r = range();
     const d = r[1] - r[0];
-    const p = d * 0.1;
+    const p = d * 0.01;
     return [r[0] - p, r[1] + p];
   };
 
--- a/src/pages/household/output/EarningsVariation/BaselineAndReformChart.jsx
+++ b/src/pages/household/output/EarningsVariation/BaselineAndReformChart.jsx
@@ -148,17 +148,9 @@ function BaselineAndReformTogetherPlot(props) {
     baselineValue,
     useHoverCard = false,
   } = props;
-  const x1 = earningsArray;
-  const y1 = baselineArray;
-  const y2 = reformArray;
-  const x2 = [currentEarnings];
-  const y3 = [currentValue];
-  const y4 = [baselineValue];
-  const xaxisValues = x1.concat(x2);
-  const yaxisValues = y1.concat(y2, y3, y4);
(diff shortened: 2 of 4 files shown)


abhcs:
While this is an acceptable solution, I think we may be able to do better by using `yref`. Hence I am converting the PR to draft.

abhcs:
Ideally, we should use Plotly *shapes* and not *plots* for cliffs because they are just annotations. Shapes also support "paper" coordinates using `yref`, for instance. This is what we want for the y-coordinates of the cliff. However, Plotly shapes do not support hover text at the moment, as far as I can tell. Therefore, I am still passing the y-axis range to the `getCliffs` function, as mentioned in the PR description.

MaxGhenis:
This looks a bit strange to me going negative. Could you bound the x and y axes at zero in cases where the data doesn't contain negative values? We'd like to do that regardless.

abhcs:
Are you sure you want zero padding when there are only nonnegative values? If I make it zero then we get:

Screenshot 2023-12-21 at 2 26 10 PM

The zeros are colliding. We can also do a smaller padding when there are only nonnegative values:

Screenshot 2023-12-21 at 2 29 33 PM

Which image do you prefer: top or bottom?

abhcs:
We can also reduce the padding everywhere from `10%` to `1%` of the length of the range of values, which would yield:

Screenshot 2023-12-21 at 2 34 52 PM

Another option to consider.

MaxGhenis:
your most recent suggestion looks good, thanks

MaxGhenis:
Ah actually could we set the y axis minimum to zero? The cliff shading extends beyond the x axis as is. We can keep the 1% margin for the x axis minimum so the dot remains visible (baseline net income is usually >0 so shouldn't be an issue for the other).

abhcs:
I can ensure that the cliff starts at zero in the above image if that is the only objection and leave the padding unchanged.

abhcs:
After latest commit:

Screenshot 2023-12-21 at 6 22 24 PM

MaxGhenis (approved):

PR #982, opened 2023-12-18
Show deltas by default on Earnings Variation and Marginal Tax Rate charts when a reform is present

### LINKED ISSUE #233: Show deltas by default on net income and MTR charts when a reform is present
I think this is the more common use case

abhcs:
Fix #233.

Tested on a random household and policy to verify that difference is displayed by default:

Screenshot 2023-12-18 at 2 34 12 PM

Screenshot 2023-12-18 at 2 34 43 PM


anth-volk (approved):
Thanks so much for this work, @abhcs! It's a long-standing issue that's great to clear up.

PR #939, opened 2023-12-05
Fix y-axis format issues in charts such as the UK NI LEL parameter chart

abhcs:
## Description

Fixed #854 by changing the calculation of the axis format in `getPlotlyAxisFormat`. This pull request also has implications for #425 and #66: the format issues for MTR should go away and the range of the y-axis should rarely exceed 100%.

## Method for fix

* Change `getPlotlyAxisFormat`
  * Remove unnecessary arguments.
  * Calculate the number of fractional digits to be displayed in a more sophisticated manner.
* Display the variable label in `EarningsVariation` chart's title and description.
* Fix the `MarginalTaxRates` description.
* In `EarningsVariation` and `MarginalTaxRates` charts,
  * Do not forcibly set precision using `tickformat`.
  * Instead depend on passing all `x` and `y` values to `getPlotlyAxisFormat`.

## Tests

I have checked that the axes in `EarningsVariation`, `MarginalTaxRates`, and `ParameterOverTime` charts are displayed properly.

Screenshot 2023-12-07 at 9 09 18 PM

Screenshot 2023-12-07 at 9 11 34 PM

Screenshot 2023-12-07 at 9 16 47 PM

### Summary
📊💰🧹

This pull request improves the data formatting and presentation of various plotly charts in the app. It refactors the code to use a common function `getPlotlyAxisFormat` for determining the best axis format for different types of variables. It also makes some minor style and naming changes, and makes the `ResultsPanel` component more flexible for different variables.

> _The `ResultsPanel` got a makeover_
> _With dynamic titles and axes to show for_
> _The data was refactored_
> _The style was adapted_
> _And the code became cleaner and lower_

### Walkthrough
*  Refactor `getPlotlyAxisFormat` function to handle different currency units, precision, and range based on the values passed as an argument, and to handle boolean or binary variables ([link](https://github.com/PolicyEngine/policyengine-app/pull/939/files?diff=unified&w=0#diff-b5951a72ed1475216756f45c9a54ea34fb15cce384284eb48b32a30c5c0bf58bL220-R257), [link](https://github.com/PolicyEngine/policyengine-app/pull/939/files?diff=unified&w=0#diff-b5951a72ed1475216756f45c9a54ea34fb15cce384284eb48b32a30c5c0bf58bL295-R282))
[...]

PR #988, opened 2023-12-19
Fix display for parameter over time chart for units such as years, kWh, etc.

### LINKED ISSUE #987: The parameter over time chart is not visible for some parameters
[NY CTC minimum age](https://policyengine.org/us/policy?focus=gov.states.ny.tax.income.credits.ctc.minimum_age&region=us&timePeriod=2023&baseline=2&reform=2), for instance:

image

abhcs:
## Description

- We fix #987.
- We display units `yrs` and `kWh` on the y-axis.

## Changes

- This regression described in #987 was introduced in #939 which used a direct call to `getPlotlyAxisFormat` instead of wrapping it in the spread operator. Before #939, the `yaxis` property was set to equal `{...getPlotlyAxisFormat(...)}` but in #939 this was changed to `getPlotlyAxisFormat(...)`, i.e., the spread operator was removed. However, `getPlotlyAxisFormat` returns `undefined` in many cases. Using the spread operator results in `yaxis = {...undefined} = {}`, and not using the spread operator results in `yaxis = undefined`. In the latter case, plotly does not display the chart. Therefore, we use the spread operator in this PR.
- We improve the `getPlotlyAxisFormat` function so that it handles the units `years` and `kilowatt-hour`. Thus, it returns `undefined` in fewer cases. There are still some unhandled units such as `list`, `variable`, and `program`.

## Screenshots

Screenshot 2023-12-19 at 11 03 19 AM
Screenshot 2023-12-19 at 11 03 43 AM
Screenshot 2023-12-19 at 11 04 15 AM
Screenshot 2023-12-19 at 11 05 09 AM

## Tests

We checked that the parameter over time chart was displayed with the correct y-axis for currencies, percentages, speeds, and ages, as shown in the above screenshots.

## Suggestions

- It would help if the parameters also contained a data type property, e.g., `int`, `float`, etc.

### DIFF
--- a/src/api/variables.js
+++ b/src/api/variables.js
@@ -242,17 +242,18 @@ export function getPlotlyAxisFormat(unit, values, precisionOverride) {
     return d > 10 ? 0 : d === 0 ? 1 : 1 - Math.round(Math.log10(d));
   };
 
-  if (Object.keys(currencyMap).includes(unit)) {
+  const isCurrency = Object.keys(currencyMap).includes(unit);
+  const isPercent = unit === "/1";
+  const isYears = unit === "years";
+  const isKwh = unit === "kilowatt-hour";
+  const isNumber = isCurrency || isPercent || isYears || isKwh;
+  // TODO: unhandled units: list, variable, program
+  if (isNumber) {
     return {
-      tickformat: `,.${precision()}f`,
-      tickprefix: currencyMap[unit],
-      ...(values && {
-        range: paddedRange(),
-      }),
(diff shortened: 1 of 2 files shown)


anth-volk (approved):
Thanks for your contribution, @abhcs, this is great work. I'll open a new issue for handling chart units more consistently, and you're right, perhaps there should be a change in the underlying model to enable this kind of thing automatically. I did make one minor change, adding a space between the number and the label.

PR #992, opened 2023-12-20
Add syntax highlighting to `Reproduce in Python` code

### LINKED ISSUE #168: Color-code `Reproduce in Python` code

abhcs:
## Description

We fix #168, i.e., we add syntax highlighting to the `Reproduce in Python` in the household and policy calculators. Since the GPT prompt in `Analysis` also used the same component, it is displayed closer to the theme in our new component but without syntax highlighting because prompts are just natural language.

## Changes

- Refactor the Python code generation and display logic into separate files. The new display component is called `CodeBlock`.
- We use [prism-react-renderer](https://github.com/FormidableLabs/prism-react-renderer) in `CodeBlock`.
  - We used some code from [this official sandbox](https://codesandbox.io/p/sandbox/prism-react-renderer-example-u6vhk?file=%2Fsrc%2Fstyles.js).
- The policy and household calculators use `CodeBlock` with language `"python"`. The `Analysis` component uses `CodeBlock` with language `"markdown"` (this is the most natural language supported by `prism` fit for GPT prompt but we are not ruling out better [options](https://github.com/FormidableLabs/prism-react-renderer#language)). 

## Screenshots

Policy:
Screenshot 2023-12-20 at 12 38 51 AM

Household:
Screenshot 2023-12-20 at 12 37 07 AM

Analysis:
Screenshot 2023-12-20 at 12 38 31 AM

## Tests

We checked that the code was displayed properly for a couple of policy and household configurations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "moment": "^2.29.4",
         "openai": "^4.20.1",
         "plotly.js": "^2.27.1",
+        "prism-react-renderer": "^2.3.1",
         "react": "^18.2.0",
         "react-bootstrap": "^2.9.1",
         "react-detect-print": "^0.1.2",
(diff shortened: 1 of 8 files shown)


MaxGhenis:
Can you avoid the extra space from lines 10+ vs <10?

anth-volk:
One last item: could you expand the width to fill the container in the two relevant pages? At present, the code sits at a default width that's a bit narrow. Otherwise, thanks for your contributions on this and many others, @abhcs.

abhcs:
@anth-volk Please check if it is better now. Since it is scrollable content, I am not posting screenshots.

anth-volk (approved):
This is perfect, thanks so much for this! Makes it look much more normal. Also appreciate the refactoring into a separate component.

Issue #218, opened 2023-02-08
Let users select a year in the household calculator


MaxGhenis:
This is especially valuable now as users may want to use PolicyEngine to estimate 2022 taxes. We only allow 2023 inputs.

A campaign to help people claim the Washington Working Families Tax Credit prompted this.

MaxGhenis:
Let's prioritize this, given more users are looking to 2024 simulations this month, and those looking for help with 2023 taxes will want it starting in January.

anth-volk:
Looking at mocking this up early next week. Which do you think is preferable: a complete page in the household flow to select a year, or a method by which a user would opt out of the current year?

MaxGhenis:
The latter - let's default to current year

anth-volk:
For clarification, we're looking for this for both the household and policy reform calculators, right?

anth-volk:
Also, as a side note, I'm looking to solve this after #1017, as solving and pushing that will make this simpler

anth-volk:
I think one option (which fits more with these specifications) would be to do something like this, which is just roughly worked up here:
<img width="1440" alt="Screen Shot 2023-12-26 at 3 28 53 PM" src="https://github.com/PolicyEngine/policyengine-app/assets/14987227/167bbe37-7745-4e30-bc7e-7c0e346e3999">

Another is to build a separate "year" input screen before all the other ones. The latter feels more intuitive, given the flow, but the former can be done. I just think it wouldn't be prudent to put it on the right-hand panel, akin to the policy reform calculator, as that feels unintuitive from a user perspective.

MaxGhenis:
Either works as part of the questionnaire - let's start by only allowing 2021-2024

MaxGhenis:
Congress is reportedly exploring retroactive reforms to 2023 tax rules, in particular the Child Tax Credit. So this just became more important, as it would allow us to show the impacts to specific households. A key policy parameter changed from 2023 to 2024 (the Additional Child Tax Credit rose from $1,600 to $1,700) so this is pretty critical--policy conversations are moving fast so let's make this a top priority.

anth-volk:
For sure. #1145 should enable this feature, it just needs added tests, which requires the merging of a different PR that I mentioned in the comments #1145, followed by the actual test writing and fixes.

PR #1022, opened 2023-12-26
Remove hardcoding of year

### LINKED ISSUE #1017: Avoid hard-coding the current year
Many files hard-code 2023. We should centralize it as a constant instead.

anth-volk:
## Description

Fixes #1017. This removes the hardcoding of the year that we currently maintain in the application, especially in the household input. It does not allow for selection of the year, which will come via a different PR and may change some of this code. This is necessary before beginning work on #218.

## Changes

These commits create a new `constants.js` file that defines three things: a `defaultYear` (fetched via `new Date().getFullYear()`), a `defaultStartDate` (a string that appends `-01-01` to the year) and a `defaultEndDate` (the same, but with `-12-31`). This component is then utilized in place of all relevant references to 2023 except the `updateHousehold` function within `HouseholdPage.jsx`.

`updateHousehold`, meanwhile, removes unnecessary data from back-end households created before September 2023. Because the household may not use 2023 as its year, to avoid runtime issues, the component first checks what year the household contains, then uses that value to remove unnecessary values.

## Screenshots

Loom video available [here](https://www.loom.com/share/fcfc76f5b9aa46adac2a6226e5d69271?sid=19732957-4948-4ee3-8cbf-117fb70ce758). Note that nothing has changed visually.

### DIFF
--- a/src/api/variables.js
+++ b/src/api/variables.js
@@ -1,6 +1,7 @@
 import { countryApiCall } from "./call";
 import { capitalize } from "./language";
 import { defaultHouseholds } from "../data/defaultHouseholds";
+import { defaultYear } from "data/constants";
 
 export function removePerson(situation, name) {
   // Remove a person from the situation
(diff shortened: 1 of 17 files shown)


anth-volk:
Note that `policyengine-api` #[1030](https://github.com/PolicyEngine/policyengine-api/issues/1030) must be resolved before this PR can be integrated.

nikhilwoodruff (approved):

PR #1013, opened 2023-12-25
Add a third `% difference` option to household net income chart

MaxGhenis:
Fixes #997

Fixes #1016

PR #1024, opened 2023-12-27
Ignore the build directory for `eslint` and restore `eslint` in `npm run fix`

### LINKED ISSUE #1021: Ignore the build directory for `eslint` and restore `eslint` in `npm run fix`
Removed in #1013 

Per https://github.com/PolicyEngine/policyengine-app/issues/1016#issuecomment-1869646241 we may be missing some fixes without it

The workflow that causes the error is as follows: run `make build` and then run `npm run lint` or `npm run fix`. eslint reports errors inside the build dir, which it should ignore.

abhcs:
## Description

Fix #1021. The `eslint` subcommand in `npm run lint` or `npm run fix` commands was reporting too many errors because they were reading the build folder created by `make build`. The `eslint` subcommand was removed in #1013 because then we did not know that the build folder was causing the problem. Now, we restore the `eslint` subcommand in `npm run fix` and add the build folder to .eslintignore.

## Tests

Tested before and after adding .eslintignore to see whether the change was working.


anth-volk:
@abhcs Thanks for digging into the mundane world of linting. This looks good, but I can't reproduce, since I have no build directory locally. @MaxGhenis Could you run, make sure this works in whatever environment yours failed in, then we can push?

MaxGhenis (approved):
I checked this out and verified that `npm run fix` works after running `make build`

PR #1061, opened 2023-12-29
fix: changed the default state to california

### LINKED ISSUE #1057: Change default state display to mirror actual default state
At the moment, when inputting a household, the state page displays a default state of "Alabama," but the actual default state is California. This should be changed, as it's very confusing if you're just clicking through.

dotslashbit:
## Description

closes #1057 

## Changes

changed the default state to california

## Screenshots

https://github.com/PolicyEngine/policyengine-app/assets/46224156/22bc6d7a-b001-4068-bf22-6b06be3929ef

## Tests

### DIFF
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -210,19 +210,25 @@ function HouseholdVariableEntityInput(props) {
         metadata,
       )
     : null;
-  let defaultValue =
-    reformValue !== null
-      ? reformValue
-      : inputValue !== null
-        ? inputValue
-        : simulatedValue;
+  let defaultValue = null;
+
+  if (reformValue !== null) {
+    defaultValue = reformValue;
+  } else if (inputValue !== null) {
+    defaultValue = inputValue;
+  } else if (simulatedValue !== null) {
+    defaultValue = simulatedValue;
+  } else if (variable.defaultValue !== undefined) {
+    defaultValue = variable.defaultValue;
+  }
+
   if (defaultValue === null) {
     if (variable.valueType === "float" || variable.valueType === "int") {
       defaultValue = 0;
     } else if (variable.valueType === "bool") {
       defaultValue = false;
     } else if (variable.valueType === "Enum") {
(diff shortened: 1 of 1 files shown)


abhcs:
The variable editor is for all variables, not just states. We should add a check that the variable is the state variable.

Maybe searching for California is better than hardcoding `2`.

@anth-volk If California is the default state, is this fact present in the model, e.g., `metadata`? If yes, then we should use the  default state in the model; otherwise, we should add this fact to the model.

MaxGhenis:
policyengine-us sets California as the default state:

https://github.com/PolicyEngine/policyengine-us/blob/36394152bcf2820c4138a06e6c55d9e9cc78ca02/policyengine_us/variables/household/demographic/geographic/state_code.py#L68-L71

abhcs:
@MaxGhenis Thanks for that information.

The problem should be fixed by improving the logic for extracting the default value from the model on lines 190-218. It is only when no defaults have been found that the first value is chosen by default on line 225.

@dotslashbit Do you want to try to find what goes wrong on lines 190-218, i.e., why the default state is not found there, and what the code should do instead?

dotslashbit:
@abhcs I have improved the logic and now its getting the default state as california, please let me know if there's any correction needed, thanks for letting me work on this.

dotslashbit:
https://github.com/PolicyEngine/policyengine-app/assets/46224156/05b8f8f0-63ef-4350-bba0-51849d5e68d1

abhcs:
@dotslashbit so we were not reading the default variable value at all -- good catch! I think these changes are good.

Looking at the code I still think that the process for choosing the default value is overly complex. Why does it have to look through 4 options to get the default? It should just be 2. Perhaps the design of these components should be improved (not in this PR).

dotslashbit:
@abhcs Thank you, so is it good to get merged?

abhcs:
@dotslashbit I'm not authorized to merge requests but someone with authority will help you out in due time.

anth-volk (approved):
Thanks @dotslashbit, just tested and this works great.

PR #1069, opened 2023-12-30
Fix bugs in titles and hover messages for policy output charts

### LINKED ISSUE #1037: Change `benefits` to `increases the net income of` in intra-decile chart
These would be more neutral/objective, and match the language in other decile plots

### LINKED ISSUE #1038: Numbers in title are localized according to the browser, not the region in the url
This is what I see when I open [this chart](https://policyengine.org/us/policy?focus=policyOutput.intraDecileImpact&reform=3170&region=us&timePeriod=2023&baseline=2) with my browser's geolocation set to Mumbai:

Screenshot 2023-12-27 at 5 22 48 PM

The number in the title is localized to the browser's geolocation. Other numbers are localized to en-US since we are viewing policyengine.org/us.

Expected: The output should be consistent: all numbers should be localized to en-US or Mumbai. The former is preferable.

### LINKED ISSUE #1064: Labor supply chart titles trillions as `kbn`
image

https://policyengine.org/us/policy?focus=policyOutput.laborSupplyImpact&reform=43238&region=enhanced_us&timePeriod=2024&baseline=2

abhcs:
## Issues

- Fix #1037.
- Fix #1038.
- Fix #1064.
- Fix other issues listed below

## Bugs

- For all charts, some numbers in titles and hover messages are localized according to the geolocation of the browser.
- In `AverageImpactByDecile`, the hover message uses relative change logic (see lines 120-140) but the change is absolute, not relative.
- In `BudgetaryImpact`'s title, the `no change` case is not considered.
- In `CliffImpact`,
  - the `y`-values are rounded (see definitions for `cliffShareChange` and `cliffGapChange`) but the hover messages are displaying the original unrounded values, and
  - `cliffShareChange` is rounded to the nearest integer, instead of two decimal places as expected
- In `DeepPovertyImpactByGender`, the title says `poverty` instead of `deep poverty`.
- In `DetailedBudgetaryImpact`
  - there are no hover messages, and
  - the style of the connectors in the chart does not match `BudgetaryImpact`.
- In `InequalityImpact`, the hover messages say `increase/decrease` twice. Other charts convey the change more efficiently.
- In `IntraDecileImpact` and `IntraWealthDecileImpact`,
  - the title only talks about gains but the chart shows both gains and losses, and
  - the case `useHoverCard = false` is not working due to the changes made when adding legends to the chart.
- In `LaborSupplyResponse`,
  - the number in the title is hardcoded to contain `bn` instead of using the abbreviation function `aggregateCurrency` properly, and
  - the title does not consider the `no change` case

## Style issues

- In all charts, the hover code is repeated twice for the two possible `useHoverCard` values.
- A lot of code was repeated in `RelativeImpactByWealthDecile` from `RelativeImpactByDecile`.

## Changes

- In `language.js`, we replace functions such as `aggregateCurrency` with functions such as `formatCurrencyAbbr` that use the country id in the metadata. We use these new functions from all impact charts.
- For all charts,
  - we add the locale code to plotly config, and
  - we add a region phrase to the title.
- We write new functions `relativeChangeMessage` and `absoluteChangeMessage` that help with creating hover messages for all charts. They help in standardizing the hover message by providing templates for the message for all charts.
- In `AverageImpactByDecile`, we use absolute change logic in the hover message.
- In `BudgetaryImpact`'s title, we take the `no change` case into account.
- In the cliff impact chart,
  - we use the rounded values (`cliffShareChange` and `cliffGapChange`) in the hover messages, and
  - we fix the rounding logic for `cliffShareChange`.
- We fix the title in `DeepPovertyImpactByGender` to say `deep poverty` instead of `poverty`.
- In `DetailedBudgetaryImpact`,
  - we add hover messages similar to `BudgetaryImpact`, and
  - we change the style of the connectors in the chart to match `BudgetaryImpact`.
- Change the messages in `InequalityImpact` to match other charts.
- In `IntraDecileImpact` and `IntraWealthDecileImpact`,
  - change the title to talk about both gains and losses, and
  - remove the `useHoverCard = false` case.
- In `LaborSupplyResponse`,
  - we remove the `bn` term hardcoded in the title and use `formatCurrencyAbbr` instead, and
  - we take the `no change` case into account.
- In all charts, we extract the hover message logic to a new function called `hoverMessage` and call it from the two locations. This improves the style of the code by reducing unnecessary repetition.
- We reuse code from `RelativeImpactByDecile` in `RelativeImpactByWealthDecile`.

## Tests

We tested all policy output charts for a few configurations, including when the browser's geolocation was set to Mumbai.

## Screenshots

Screenshot 2023-12-30 at 7 39 05 AM
Screenshot 2023-12-30 at 7 44 40 AM
Screenshot 2023-12-30 at 7 47 52 AM
Screenshot 2023-12-30 at 8 13 55 AM


MaxGhenis:
Looks great, how about this (pseudocode) to shrink down the title for the intradecile chart? (note it's the share of population not households)

```
if winners > 0:
    if losers > 0:
        {reform} increases net income for x% of the population, and decreases it for y%.
    else:
        {reform} increases net income for x% of the population.
else:
    if losers > 0:
        {reform} decreases net income for x% of the population.
    else:
        {reform} does not affect anyone's net income.
```

abhcs:
@MaxGhenis I have used your pseudocode but have tried to keep the text consistent with other charts. For instance, in this PR, I am using phrases such as `reform would increase` and `reform would have no effect on` everywhere.

Screenshot 2023-12-30 at 10 05 49 AM

MaxGhenis:
Sounds good!

I see you've improved the rounding. While we're doing that, could we add a trillions option? For example in the chart you showed above, could we say `Reform would increase employment income by $1.09tn this year` and similarly use that in the labels for each bar

abhcs:
Added `tn` and the `-` sign is before the currency symbol even for currency abbreviations:

Screenshot 2023-12-30 at 10 30 14 AM

MaxGhenis (approved):

PR #1081, opened 2023-12-31
Configure linter properly for react and fix errors in current code

### LINKED ISSUE #1080: Configure linter properly for react and fix errors in current code
eslint is not configured properly for react. Consider [this example](https://react.dev/learn/lifecycle-of-reactive-effects#react-verifies-that-you-specified-every-reactive-value-as-a-dependency) from react docs that a linter should be able to detect. When I add this example to our src directory, it is unable to detect the error.

On configuring eslint properly, as recommended [here](https://react.dev/learn/editor-setup#linting), I get 50 errors and 36 warnings in the src directory. A portion of the errors is shown below:

Screenshot 2023-12-31 at 11 34 33 AM

This means that we are not complying with the minimum recommendations from react on how to use hooks and other features. We should
- fix these errors, and
- configure eslint properly

abhcs:
We fix #1080 by enabling basic eslint rules and fixing errors found by the linter.


abhcs:
Now that we have fixed the react hook errors, let us enable the react-app config in eslint.

abhcs:
After each commit, I tested that the affected functions in the app were working as expected. And now the code passes lint for the recommended config for react, i.e., `react-app`.

anth-volk:
Before checking this, I want to confirm with @nikhilwoodruff or @MaxGhenis - I wouldn't be surprised if merging this created merge conflicts down the road. Obviously, this seems quite important, but are there any features we're currently working on that should be complete before reviewing and merging?

abhcs:
Please review this sooner rather than later. 1000+ lines are affected in this PR only because of the deletion of unreachable code (I didn't want to fix errors in dead code) -- it won't create too many conflicts. If delayed, this PR will get harder to merge because we may keep adding code to master that does not pass eslint's `react-app` config, which as mentioned in the issue description is quite basic and used to be enabled in the past. It is better to merge the PR sooner so that contributors can detect more errors in their contributions.

This PR has higher priority than #1098, which only affects a few users who have enabled `husky`.

Please take your time with the review though -- I don't wish for incorrect code to get merged.

src/pages/household/input/VariableEditor.jsx
```diff
@@ -47,13 +38,15 @@ export default function VariableEditor(props) {
...
   // Add the variable to the relevant portions of the household input object
   useEffect(() => {
-    const newHouseholdInput = addVariable(
-      householdInput,
-      variable,
-      entityPlural,
-    );
-    setHouseholdInput(newHouseholdInput);
-  }, [variable]);
+    if (!possibleEntities.length) {
```
nikhilwoodruff, line 41: Note to test this doesn't break anything

nikhilwoodruff (approved):
Thanks- LGTM!

PR #1104, opened 2024-01-04
Fix error encountered in deployment

abhcs:
Fix lint error that prevented [deployment](https://github.com/PolicyEngine/policyengine-app/actions/runs/7413645656/job/20173038901).

The lint error was created in the master branch after PR #1080 was created -- therefore, it was not caught. We should not see these errors during deployment after merging new PRs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "fix": "eslint --fix --ext js,jsx . && prettier -w .",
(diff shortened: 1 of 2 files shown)


abhcs:
The deployment error says:

> Treating warnings as errors because process.env.CI = true.
> Most CI servers set it automatically.

Since warnings are treated as errors during deployment, we should use the same setting in the PR workflow.

anth-volk (approved):
All seems logical to me, thanks @abhcs.

PR #1115, opened 2024-01-05
arrow buttons hover color change disabled

### LINKED ISSUE #1109: Buttons without words on them should not allow highlighting
For example, the arrow buttons on `BottomCarousel` (the component at the bottom of the screen when you calculate a household) allows the user to highlight the arrows themselves; this shouldn't be the case.

dotslashbit:
## Description
fixes #1109  Removed the on hover transition and color change on the arrow buttons
## Screenshots

https://github.com/PolicyEngine/policyengine-app/assets/46224156/351f4a5a-8b5b-4bdc-b200-bb0254068641

### DIFF
--- a/src/controls/ArrowButton.jsx
+++ b/src/controls/ArrowButton.jsx
@@ -1,5 +1,5 @@
 import Button from "./Button";
-
+import styles from "../redesign/style";
 export default function ArrowButton(props) {
   const { direction, onClick, type, style } = props;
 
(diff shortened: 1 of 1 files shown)


anth-volk:
Thanks so much for this @dotslashbit, but unfortunately this isn't quite what we were looking for. Perhaps the original issue is ambiguous, but what I meant is that the user is able to drag their cursor over the arrow and highlight it as if it were text, akin to when you highlight some text in Microsoft Word or something like that. We would like to maintain the hover transition, just not allow the user to highlight the arrow itself. 

If you'd like me to clarify more, I can send a video, if you'd like.

Closing.

dotslashbit:
If this is still an issue, then please clarify this and I'll work on it, thanks

anth-volk:
I'll provide details in the issue itself

PR #1124, opened 2024-01-07
Repair Jest

### LINKED ISSUE #1123: Have Jest correctly import ES6 components
When writing tests, it's currently impossible to test a particular subset of components because we run into an error whereby Jest is unable to import ES6 modules. For example, it's impossible to test the broader functionality of the mobile policy page because, somewhere in its import tree, it or something it relies upon attempts to import `react-markdown`, which breaks as Jest is unable to handle modules. Fixing this would be vital to enabling regression testing, as well as standard testing for some components and integration testing that relies upon manipulating visual components.

anth-volk:
## Description

Fixes #1123. This PR re-enables proper Jest testing of ES6 modules, then fixes a variety of issues created by implementing its solution method.

## Changes

At present, we utilize the `react-scripts test` command to execute tests. This command invokes Jest with a variety of React-friendly presets. However, it inherently limits our ability to customize our implementation of Jest; for example, react-scripts is unable to handle certain custom presets. Indeed, our current implementation actually ignores both our preexisting Jest config file (`jest.config.js`) and our defined Jest settings within `package.json`. As a result, this PR replaces `react-scripts test` with `jest`, running standard Jest outside of the `react-scripts` environment without requiring us to eject from `react-scripts`.

As a result of this, though, this PR is also forced to re-enable various features that `react-scripts` does behind the scenes. It installs Babel to convert ES6 modules into CommonJS modules to allow processing via Jest. It polyfills `fetch` with `node-fetch` (when used within the test itself) and `whatwg-fetch` (when an imported component is using fetch). It also globalizes an encoder and decoder and adds a reset for an object that was causing the `InputField` test to fail post-change. Finally, it mocks image and CSS imports, as Jest is unable to process these natively.

## Screenshots

Unnecessary, as tests are built into the PR GitHub Action flow.

## Tests

N/A


abhcs:
Jest supports ESM through `--experimental-vm-modules`: https://jestjs.io/docs/ecmascript-modules. Did you try this flag?

anth-volk:
> Jest supports ESM through `--experimental-vm-modules`: https://jestjs.io/docs/ecmascript-modules. Did you try this flag?

Yes, I did. If trying to do so while using `react-scripts`, either the flag won't pass through or, if you're able to pass it by adding `-- ` before the flag, it completely breaks. When using it manually through Jest, we still run into a variety of issues, as well.

nikhilwoodruff (approved):
LGTM after removing excess console.log statements

PR #1128, opened 2024-01-08
fix: document title now updating properly

### LINKED ISSUE #1119: Tab title is `Current law (compared against current law)` for no-reform household simulations
e.g. https://policyengine.org/us/household?focus=householdOutput.earnings&household=39237

dotslashbit:
## Description

closes #1119 

## Changes
Used useEffect to update document title

## Screenshots
Screenshot 2024-01-08 at 2 10 43 PM

### DIFF
--- a/src/pages/household/output/HouseholdOutput.jsx
+++ b/src/pages/household/output/HouseholdOutput.jsx
@@ -9,9 +9,10 @@ import MarginalTaxRates from "./MarginalTaxRates";
 import NetIncomeBreakdown from "./NetIncomeBreakdown";
 import PoliciesModelledPopup from "./PoliciesModelledPopup";
 import HOUSEHOLD_OUTPUT_TREE from "./tree";
-import React, { useEffect } from "react";
+import React from "react";
 import { message } from "antd";
 import ResultActions from "layout/ResultActions";
+import { Helmet } from "react-helmet";
 
 export default function HouseholdOutput(props) {
   const [searchParams] = useSearchParams();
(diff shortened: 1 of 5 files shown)


anth-volk (changes requested):
Thanks for this, @dotslashbit. That said, I don't believe this solves the issue. What I think Max is looking for here is for, in situations where we're looking only at current law, the `policyLabel` should just be "Current law". There's an issue in the underlying logic where the app seems to think there is a reform, and "Current law" is set as both the reform and the baseline. Could you look into this a bit more and refactor?

dotslashbit:
ok, will update on this PR after refactoring

dotslashbit:
Screenshot 2024-01-11 at 11 45 19 AM

@anth-volk Let me know if there's any correction needed

anth-volk:
This appears to be getting the same README linting message that the other PR is getting. Once you paste your terminal output into that one, I'll try to figure out what's going on.

anth-volk:
@dotslashbit Thanks again for this. I'm actually going to hold off on this until I see the verdict on #1169 because I think the package that that adds (`react-helmet`) would allow far easier updating of the page titles. If #1169 is merged, I hate to do this, I'm hoping to ask you to rewrite the logic in this to instead use the `Helmet` component at the top of the relevant components and just declare a `<title>` tag to fill in the title, as that's far cleaner than what we've been doing so far.

dotslashbit:
ok, no issue

dotslashbit:
@anth-volk As #1169 has been merged, please inform what changes you want me to do, I'll do that

anth-volk:
@dotslashbit If you could keep the logic you wrote to determine the `policyLabel` variable, but where the old code manually set `document.title` in each component, if you could instead use the `react-helmet` package above the output JSX, like this:

```
<Helmet>
  <title>{title}</title>
</Helmet>
{the rest of the code}
```

I'll be honest, I'm new to the package myself, so if you run into any issues, please feel free to check into their docs [here](https://www.npmjs.com/package/react-helmet) or to write again in this thread.

dotslashbit:
Screenshot 2024-01-13 at 2 44 32 AM

dotslashbit:
@anth-volk I have updated the code and its working as expected, please take a look and let me know if there's anything more to be done, thanks

src/pages/household/output/HouseholdOutput.jsx
```diff
@@ -37,12 +38,26 @@ export default function HouseholdOutput(props) {
...
+    policyLabel = "Current law";
+  } else if (reformLabel !== "Current law" && baselineLabel === "Current law") {
     policyLabel = reformLabel;
   } else {
     policyLabel = `${reformLabel} (compared against ${baselineLabel})`;
   }
   let pane;
+  let title;
+  if (policyLabel) {
+    if (focus === "householdOutput.netIncome") {
```
anth-volk, line 51: Previously, each sub-component within the code managed its own title. Now that we've enabled helmet, can you return that logic so that each component can be more self-contained and we avoid this nested if?

anth-volk (changes requested):

anth-volk (approved):
This looks great! Thanks so much @dotslashbit.

PR #1130, opened 2024-01-08
Household questionnaire asks for head age when multiple members

### LINKED ISSUE #1129: Household questionnaire doesn't ask for head age when multiple members
See this video: first scenario works (single no kids), all others drop the head when asking for age

https://www.loom.com/share/b0051eea508b4c5296e0cd4c09779fc3

abhcs:
## Description

Fixed #1129. The regression was introduced in #1081 when I tried to fix the `react-hooks/exhaustive-deps` error in the `VariableEditor` component.

Before #1081, `variable` was listed as a dependency so that the effect would execute once for every variable and the dependency list was not exhaustive. Leaving aside the `exhaustive-deps` error, the dependency was also inefficient -- passing large objects as dependencies is an anti-pattern. Using just `variable.name` is more efficient but it does not get around the `exhaustive-deps` error.

From the documentation of `addVariable` I gathered that the purpose of the effect was to add the variable to all entities. `!possibleEntities.length` is true if and only if the variable is absent in all entities. Therefore, I passed no dependency array at all and added the condition `!possibleEntities.length` to the effect in #1081.

There is a subtle error in the reasoning in the above paragraph. I did not know then that a variable (such as `age`) may be present in some entities, and `addVariable` only adds the variable to the entities where it is missing. So, the effect needs to run whenever `possibleEntities.length !== householdInput[entityPlural].length`, i.e., the variable is absent in *some* entities (not *all*, as in previous paragraph).

The application logic in the front end for households is quite complex. This complexity is reflected in `addVariable` (see nested conditions) and the use of `addVariable` in the effect. Some of the default values of age are being fetched from `variable.defaultValue` while others are simply stored as constants in the front end, e.g., in `src/data/countChildrenVars.js`. Simplification is needed here for maintenance.

## Tests

Captured some tests via [loom](https://www.loom.com/share/a2e3911607ae41338a1baa5629cb42be?sid=c00e74b4-5e49-4b3d-beae-bd166277016e).

### DIFF
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -34,12 +34,15 @@ export default function VariableEditor(props) {
   const entityPlural = metadata.entities[variable.entity].plural;
   const isSimulated = !variable.isInputVariable;
   const possibleEntities = Object.keys(householdInput[entityPlural]).filter(
-    (entity) => householdInput[entityPlural][entity][variable.name],
+    (entity) => householdInput[entityPlural][entity][variableName],
   );
 
-  // Add the variable to the relevant portions of the household input object
+  // The variable must be present in all entities. The following effect is
+  // executed whenever the variable is missing from some entities in the
+  // household. addVariable is called to ensure that the variable is added to
+  // all entities.
   useEffect(() => {
-    if (!possibleEntities.length) {
+    if (possibleEntities.length !== householdInput[entityPlural].length) {
       const newHouseholdInput = addVariable(
         householdInput,
         variable,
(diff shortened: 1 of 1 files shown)


anth-volk (approved):
Thanks for this, @abhcs, the revision is an important one.

You make a fair point regarding the complexity of the variable input components. The challenge has been that certain variables don't lend themselves to easy categorization: for example, were we to set `age` off of the API, the dependents' ages would be the same as whatever has been coded for heads-of-household. I think you're definitely correct, though.

abhcs:
You need defaults for `(entity, variable)` combinations or maybe `(entityType, variable)` combinations, just storing the default with the `variable` is not enough.

anth-volk:
Perhaps, but doing so would require rewriting the data entries for something like 10,000 variables across 6 country packages, hence why it hasn't been done

abhcs:
Are the 10,000 variables edited via `VariableEditor`? I thought it edits only about a dozen or two.

anth-volk:
They aren't all at the moment, of course, but there were times where more were editable. Either way, it'd be great to have consistency among all variables; after all, a lack of it is what created #1131.

PR #1138, opened 2024-01-09
Addendum to PR #1130

### LINKED ISSUE #1133: `VariableEditor` infinitely re-renders
Probably effect of #1081. If adding a line to console log right before component return, the fact that the component re-renders infinitely is visible.

abhcs:
Fix #1133 by fixing type error in PR #1130 (checked length of an object using `.length` instead of `Object.keys().length`).

Tested using `console.log`.


anth-volk (approved):
Thanks for this, @abhcs.

Issue #745, opened 2023-10-18
Make button highlight instant


MaxGhenis:
Hovering over the new buttons causes a fast left-to-right animation that highlights them. Modern UX paradigms seem to favor instant highlighting instead (checked Slack, GitHub, Google Workspace, Facebook, etc.)

nikhilwoodruff:
I'm opposed to this but open to argument/numbers...

MaxGhenis:
Thank you for filing this issue. Closing as part of the `policyengine-app` (v1) wind-down: policyengine.org is now served by [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2), and this repository is being archived, so this issue is frozen against a codebase that no longer deploys anywhere.

If this is still relevant on the current site, please re-file it against [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2) — we'd welcome it there. Apologies that this sat open through the migration.

PR #1147, opened 2024-01-10
fix: eliminated console warnings for PolicyPage left sidebar

### LINKED ISSUE #596: Console warning regarding animations when hovering over various interactive elements
When hovering over various different interactive elements in Firefox, I receive the following warning:

`You are trying to animate color from "rgba(0, 0, 0, 0.85)" to "white". rgba(0, 0, 0, 0.85) is not an animatable value - to enable this animation set rgba(0, 0, 0, 0.85) to a value animatable to white via the `style` property.`

Perhaps there's a Styled-Component setting in the raw CSS that is attempting to do this?

### LINKED ISSUE #597: Console warning regarding bolding when hovering over interactive elements
When hovering over various different interactive elements, I receive the following console warning:

`You are trying to animate fontWeight from "500" to "bold". 500 is not an animatable value - to enable this animation set 500 to a value animatable to bold via the `style` property.`

Perhaps there's a shared Styled-Component raw CSS setting creating this?

dotslashbit:
## Description

- closes #597 
- closes #596 

## Changes

changed fontWeight from "bold" to "700" and color from "black" to "#000"

## Screenshots

https://github.com/PolicyEngine/policyengine-app/assets/46224156/61911e29-24bf-4ec4-9456-d3e037a17146

### DIFF
--- a/src/layout/FolderPage.jsx
+++ b/src/layout/FolderPage.jsx
@@ -117,7 +117,6 @@ export default function FolderPage(props) {
                 backgroundColor: style.colors.DARK_GRAY,
                 color: "#fff",
               }}
-              transition={{ duration: 0.25 }}
               onClick={() => {
                 let newSearch = copySearchParams(searchParams);
                 newSearch.set("focus", child.name);
--- a/src/layout/FolderPage.jsx
+++ b/src/layout/FolderPage.jsx
@@ -115,7 +115,7 @@ export default function FolderPage(props) {
               whileHover={{
                 scale: 1.05,
                 backgroundColor: style.colors.DARK_GRAY,
-                color: "white",
+                color: "#fff",
               }}
               transition={{ duration: 0.25 }}
               onClick={() => {
--- a/src/layout/Menu.jsx
+++ b/src/layout/Menu.jsx
@@ -20,14 +20,14 @@ function MenuItem(props) {
       <motion.h5
         style={{ fontSize: 18 }}
         initial={{ x: 0 }}
-        whileHover={{ color: "black", fontWeight: "bold" }}
+        whileHover={{ color: "#000", fontWeight: 700 }}
       >
         {selected === name && (
           <span
             style={{
               marginRight: 10,
-              color: "black",
-              fontWeight: "bold",
+              color: "#000",
+              fontWeight: 700,
               textShadow: "0 0 .2px #000",
             }}
           >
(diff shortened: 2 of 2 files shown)


dotslashbit:
@anth-volk Please take a look and let me know if this solution is fine or not, thanks

anth-volk:
Thanks for this @dotslashbit. This does work and is what we're actually shooting for, perfectly re-enabling what we had created, but before I approve this, I have to say that this effect just looks bad. @nikhilwoodruff Can you check this out and see what you think? Should we just close this, and ask Sahil to instead put in place a different effect? It feels super sluggish.

MaxGhenis:
We should remove all delays on all effects. Modern UI has instant effects everywhere AFAICT to avoid sluggishness.

dotslashbit:
@anth-volk So, just give me clarification on what changes you want me to do and if all the changes should be in this PR or not?

dotslashbit:
@MaxGhenis I think there's already an issue on what you are talking about #745

MaxGhenis:
Right, would prefer not to augment delays in advance of that

dotslashbit:
So, what should I do, should I leave this PR as it is and create another PR and work on #745 or should I make changes in this PR itself

anth-volk:
@dotslashbit #745 is actually unrelated, as it's scoped only to buttons, which will require a more fundamental change (ripping out and replacing). For #1147, could you keep all of your changes so far and just remove the transition duration, and then test to make sure that on hover, these just automatically transition to the bold version?

anth-volk (changes requested):
This actually still does the delayed transition into expanded, bolded text. @dotslashbit Could you dig a bit more into why that is and remove it? If that means removing some of the `AnimatePresence` components and those sorts of things and merely adding a `whileHover` prop, please feel free to do so.

dotslashbit:
I left them intentionally as I thought you only want to remove animations from the links and not the cards. I'll update it now

anth-volk:
@dotslashbit You're not wrong, the animation does work as we'd actually like now, but it throws up to a hundred console warnings. The reason is because it's actually not functioning properly - it was originally intended to do the gradual animation that you re-enabled - but since that animation looks pretty bad, we're hoping you can redo it to do basically what it does now, but without throwing all the warnings.

dotslashbit:
@anth-volk ok got it, working on it

anth-volk:
@dotslashbit Sorry, I know we changed objectives a couple of times on this one

dotslashbit:
Here's the updated video for the hover effect, let me know if this is alright or this needs some more changes, thanks.

https://github.com/PolicyEngine/policyengine-app/assets/46224156/f266b5dd-d75f-4fd8-bd3f-6c996a7a33c5

anth-volk (approved):
Great, just confirmed that this works as intended. Thanks @dotslashbit!

PR #1148, opened 2024-01-10
Fixes #1143: Default date were being used which have logic of present…

### LINKED ISSUE #1143: Issue with date visibility on policy calculator
When adding a policy parameter that does not conform to the standard time period defined automatically within the policy calculator, if you revisit the parameter after calculating outputs in order to edit it, the calculator defaults to the normal default dates instead of the dates the parameter is extended over.

For example, if you visit https://policyengine.org/us/policy?focus=policyOutput.laborSupplyImpact&reform=43175&region=enhanced_us&timePeriod=2024&baseline=2, then click on "Enable Abolish federal income tax" in the policy carousel, you'll note that the parameter runs to 2028-12-31, but the input defaults to it ending on 2029-12-31. The same disparity is true for start dates.

It would be great to pre-populate based on the existing definition dates to prevent user confusion when attempting to edit a parameter.

JiginJayaprakash:
Fixes #1143: Default date were being used which have logic of present year and +5 year
because of which 2024 and 2029 were shown from https://github.com/PolicyEngine/policyengine-app/blob/master/src/data/constants.js

Updated to values coming from policy.reform.data


anth-volk (changes requested):
Thanks for this contribution @JiginJayaprakash. Unfortunately this introduces changes that break the app. At the moment, the app utilizes `defaultStartDate` and `defaultEndDate` as its first setting for the `startDate` and `endDate` stateful variables because variables that have not yet been added to the policy reform don't sit within the policy data, and thus the `Object.keys()` operation you call can't be executed. 

I think the solution to this likely rests instead in determining whether or not the parameter exists within the `policy` prop. If it does, `startDate` and `endDate` can be initialized off of that, otherwise they should be initialized off of the default values.

anth-volk:
Note to self: testing against https://policyengine.org/us/policy?focus=policyOutput.laborSupplyImpact&reform=43175&region=enhanced_us&timePeriod=2024&baseline=2

src/pages/policy/input/ParameterEditor.jsx
```diff
@@ -30,6 +31,17 @@ export default function ParameterEditor(props) {
     startDate,
   );
 
+  useEffect(() => {
+    const [parameterStartDate, parameterEndDate] = policy.reform.data[
```
anth-volk, line 35:
```suggestion
    const [parameterStartDate, parameterEndDate] = policy?.reform?.data[
```

anth-volk: This is to avoid runtime errors if, for example, `policy.reform` exists, but the `data` key does not

anth-volk, line 35: Can you refactor this effect hook to only run when `policy.reform.data[parameterName]` exists, then run your setters based upon that? At the moment, this hook re-sets `startDate` and `endDate` each time we access a new, unset parameter

src/pages/policy/input/ParameterEditor.jsx
```diff
@@ -30,6 +31,17 @@ export default function ParameterEditor(props) {
...
+      parameterName
+    ]
+      ? Object.keys(policy.reform.data[parameterName])[0].split(".")
+      : [defaultStartDate, defaultEndDate];
+    setStartDate(parameterStartDate);
+    setEndDate(parameterEndDate);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
anth-volk, line 42: Ideally, we should really avoid manually disabling `exhaustive-deps`. I know there are points in the code where it is utilized, don't get me wrong, but those are legacy and we're moving away from them. I believe that if you rewrote this to only actually update state when parameterName exists within `policy.reform.data`, you might avoid whatever issue you ran into that caused you to add the `eslint-disable-next-line` call.

src/pages/policy/input/ParameterEditor.jsx
```diff
@@ -30,6 +31,17 @@ export default function ParameterEditor(props) {
     startDate,
   );
 
+  useEffect(() => {
+    const [parameterStartDate, parameterEndDate] = policy.reform.data[
+      parameterName
+    ]
+      ? Object.keys(policy.reform.data[parameterName])[0].split(".")
```
anth-volk, line 38: For this code, can you first confirm that `policy.reform.data[parameterName]` is, in fact, an array with a length greater than 0 before attempting to split? This prevents runtime errors if, for some reason (likely a bug somewhere else), `parameterName` exists, but is empty, `undefined`, `null`, or an `Object` or something.

anth-volk (changes requested):
Thanks so much for this, @JiginJayaprakash. I want to preface by saying that, based on testing, your code does work, and it's a big help to have it. These changes just request some error strengthening and changes meant to prevent additional effect hook runs.

anth-volk:
Oh, and if you wouldn't mind rebasing off master, that would likely remove the file conflict that's showing up

anth-volk (approved):
Thanks so much for this @JiginJayaprakash, this does make the app quite a bit smoother.

I do want to note that while this improve this quite a bit, this component is far from perfect. We should fundamentally reconsider how we implement this, as even with your fixes that make the thing far better, it still has numerous strange quirks that should be dealt with in a future rewrite.

PR #1149, opened 2024-01-11
fix: pointer and link issue

### LINKED ISSUE #761: Navbar links extend beyond the visible link area
Minor issue, but the anchor tags on the navbar links extend outside of the visible link area (the area that turns lighter when hovered) into the buffer area between the nav options

dotslashbit:
## Description

closes #761 

## Changes

removed margin for each link div

## Screenshots

https://github.com/PolicyEngine/policyengine-app/assets/46224156/926446c7-a085-46da-9c4c-443a1da058c9


anth-volk:
Thanks for this @dotslashbit, I'll check this out first-thing tomorrow morning, but glancing at this cursorily, it does look good!

anth-volk (changes requested):
Thanks for this @dotslashbit. That said, I was looking into this component, and I think it needs a bit more of a refactor. While your code works, it unfortunately moves stuff over, so in order to better maintain spacing, improve the code, and accomplish what we're trying to achieve, could you do the following:

* On desktop, remove the `paddingLeft: 30` from PageLinks and instead add `gap: 45` to the overall component that's only added when `displayCategory` is desktop
* Keep the `margin: 15` removed, as you did
* Add `gap: 30` to the PageLinks to ensure they're still spaced properly

Also, please feel free in the future to refactor more thoroughly if you see weird visual effects, like using margins/padding inside of flex containers when `gap` might be more apt. Thanks so much for your contributions, @dotslashbit!

dotslashbit:
@anth-volk updated the code as required

https://github.com/PolicyEngine/policyengine-app/assets/46224156/38d64641-1259-482d-bc26-8ecadcdfbff9

anth-volk (approved):
Thanks @dotslashbit! I did take the liberty of changing one minor value, but otherwise this looks great.

PR #1151, opened 2024-01-11
1003:About page visited without countryID

### LINKED ISSUE #1003: Ensure that About page can be visited from a domain without a country ID

JiginJayaprakash:
Fixes #1003 About page can be visited without CountryID

To maintain the countryId used localstorage when on about page, and updated /about redirection


anth-volk:
Hey @JiginJayaprakash, thanks so much for this. As I was looking through the changes, honestly, I think it actually doesn't make sense to do #1003. I'm sorry, I know it might be disappointing, but I'm going to close this PR and that issue. When I opened it, I thought the other pages (Research and Donate) would also be implemented without `countryId`, but Research depends on it, and I think it'd be confusing to remove the IDs for just two pages. Even so, thanks for your efforts, they're much appreciated.

PR #1193, opened 2024-01-15
added basic button to rename and submit and shows toast if name exists

### LINKED ISSUE #972: Add prompt for duplicate policy names
When a user names a policy with a name that already exists, we should prompt them with two options:

* Choose a different name
* Un-name the old one (revert to Policy #X) and keep the current name

### LINKED ISSUE #1190: Create some sort of "rename" button that becomes "submit"
This should likely be built into #972. At the moment, it is not at all clear that a user can rename a policy. The app should make this clear by adding a button that defaults to something like "rename" as its text, but on click, allows the user to change the name and changes its text to something like "submit." This would then on click utilize basically the same handling logic that we currently have for renaming on blur and Enter keydown

dotslashbit:
## Description

- fixes #972 
- fixes #1190 

## Changes

Please describe the changes in detail.

## Screenshots

Please include concise visuals (videos, screenshots, etc.) demonstrating the code in action and the impact of changes. Many on the team use a free tool called [Loom](https://www.loom.com/) to capture and share short clips, but feel free to utilize any tool you'd like. Screenshots of the code itself can also be helpful, but are usually not necessary.

## Tests

### DIFF
--- a/src/api/parameters.js
+++ b/src/api/parameters.js
@@ -146,12 +146,14 @@ export function getNewPolicyId(countryId, newPolicyData, newPolicyLabel) {
   if (newPolicyLabel) {
     submission.label = newPolicyLabel;
   }
+  console.log("submission", submission);
   return countryApiCall(countryId, "/policy", submission, "POST")
     .then((response) => response.json())
     .then((data) => {
       if (data.status === "error") {
         return data;
       }
+      console.log("from server", data);
       return data.result.policy_id;
     });
 }
--- a/src/pages/policy/PolicySearch.jsx
+++ b/src/pages/policy/PolicySearch.jsx
@@ -1,10 +1,12 @@
-import { AutoComplete } from "antd";
+import { AutoComplete, message } from "antd";
 import { useEffect, useState } from "react";
 import { useSearchParams } from "react-router-dom";
 import { copySearchParams, countryApiCall } from "../../api/call";
 
 export default function PolicySearch(props) {
   const { metadata, target, policy, width, onSelect } = props;
+  const [isRenaming, setIsRenaming] = useState(false);
+  const [newName, setNewName] = useState("");
   const [searchParams, setSearchParams] = useSearchParams();
   const defaultId = searchParams.get(target);
   let defaultLabel = policy[target].label || `Policy #${defaultId}`;
@@ -20,6 +22,19 @@ export default function PolicySearch(props) {
   useEffect(() => {
     setValue(defaultLabel);
   }, [defaultLabel]);
+
+  const onRenameSubmit = () => {
+    countryApiCall(metadata.countryId, `/policies?query=${newName}`)
+      .then((data) => data.json())
+      .then((data) => {
+        if (data.result.length > 0) {
+          message.error("Policy name already exists");
+        } else {
+          // Handle the case where the policy name does not exist
+          // This could involve making another API call to rename the policy
+        }
+      });
+  };
 
   // The search should query the API, but limited to one request every 1000ms.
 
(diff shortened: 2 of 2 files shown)


dotslashbit:
@anth-volk While working a solution for the issue, I added a submit button that whenever a user hits that button, first it'll check if there's a policy name exists or not, if yes then show a error toast else use the same logic as PolicyNamer component.
But I think PolicyNamer doesn't work.

Current state of this PR:
Its correctly detecting if a policy exists or not and if it exists it shows a toast as Policy name already exists

My Query:
Am I right about PolicyNamer not working properly (on localhost)?

dotslashbit:
@anth-volk I may be wrong, but I think the api endpoint to submit a new policy isn't working properly. If I'm wrong or I'm using the wrong API end-point, then please let me know, thanks

dotslashbit:
@anth-volk Sorry if I'm asking and pinging too many times, but I noticed that the current implementation to rename the policy is updating the database and after that if I go to my localhost and try to rename the policy to that exact same name, then its detecting that the policy name already exists.

dotslashbit:
@anth-volk there are still changes to be made for UI. Currently my changes are showing users the prompt is there's a policy name already exists and if not it lets them create the policy with the name.
But if the user tries to rename the Policy right after the first rename, then its not working. I think you have also reported an issue regarding this.
If you can link the issue here, that would be great as I think I have to solve that issue too in this PR.

anth-volk:
@dotslashbit I should have some time late today to look into this, thanks for flagging it. I'll see what I can find.

anth-volk:
Alright, I think I figured out what's going on here. At the moment, on the live app, when a user first creates a policy, one POST request is fired to the `/policy` endpoint, creating a record with an empty `label` value and an ID number. When a user first renames their policy, the app works half-correctly. If you check out the `set_policy` endpoint [here](https://github.com/PolicyEngine/policyengine-api/blob/master/policyengine_api/endpoints/policy.py), you'll see why this is: if a new request is made and a record with the `policy_hash` and `policy_id` already exist, then the app attempts to either pull the existing label or add the new one to the overall data structure (I'll come back to that in a moment). It then creates a new record with all the same information, then updates that information (perhaps that creates the weird duplicate record we're seeing?).

However, if the user attempts to change the name another time, this outright fails, as that label pull that I mentioned before finds that the record has an existing label, so just tosses the new one out. 

I think the root of this problem is with the API–more specifically, its lack of care to follow RESTful principles. If this were RESTful, we'd use the POST request only to create records, and we'd have a separate UPDATE endpoint where we could merely update the `label` parameter, simplifying a lot of this headache, I'd think.

If you'd like, could you open an issue on the API laying all of this out? I can do one, but probably not until a bit later today, after I finish some pressing things. If not, though, I'm totally happy to do so. In the meantime, I think it'd be best to pause the work on this. 

If you want, too, you can spearhead the development of that, and I can provide assistance as you go along.

dotslashbit:
@anth-volk I'll open an issue right now but I can't work on the api issue as I'm not at all familiar with flask or python backend in general.
So, I think it would be best to leave it up to you.
For now, I'll stop working on it and keep this PR as draft.
Once the API issue is resolved, then let me know I'll continue to work on it

anth-volk:
@dotslashbit Unfortunately, as part of a different fix, I opened a PR (#1377) that would fix #1190. However, #972 is still very much open, and I'm hoping that after #1377 is merged, you might be willing to work on #972 within the new button component. I'm sorry to take a portion of your PR, but I couldn't see a way of fixing a couple other issues without rewriting the `InputField` component accordingly.

dotslashbit:
@anth-volk ok thanks, please let me know once #1377 gets merged, then I'll start working on this PR again

anth-volk:
@dotslashbit Looks like it is now. 

That said, before you do so, I'm going to merge the button code that you worked on for #1324 - thanks so much for that, it's a major improvement - but separately, the big "COMPUTE POLICY IMPACTS" button in the navbar is not a `Button` component, meaning that it won't be affected and will be stylistically different from the other buttons. Before working on this PR (#1193), could you work on changing that to match the other buttons? I'll be opening an issue for it and share below.

anth-volk:
Feel free to either rework the COMPUTE POLICY IMPACTS element to look the exact same, or to refactor it to be a `Button` element, but with the same functionality that it has now, just with different shading, etc.

dotslashbit:
@anth-volk ok, will continue to work on this PR from now on, thanks for the response

dotslashbit:
@anth-volk I'm closing this PR and will open another PR to work on #972, as most of the work in this PR is related to #1190

PR #1194, opened 2024-01-16
1191: Fix faulty policy output URLs

### LINKED ISSUE #1191: Error handle faulty policy output URLs
At present, the way a policy output/analysis component is displayed is that the search params' `focus` value is fed into a data object called `map`, then the corresponding component is displayed. However, when you try passing garbage to the URL beyond the `policyOutput.` portion (for example, https://policyengine.org/uk/policy?focus=policyOutput.test&reform=44629&region=uk&timePeriod=2024&baseline=1), the app white-screens, or if running in debug, it runtime errors. This should be properly handled by first determining if the key exists in the `map` object, and if not, displaying an error page.

JiginJayaprakash:
Fixes #1191 

Display an error message in the Error Page

### DIFF
--- a/src/pages/policy/output/Display.jsx
+++ b/src/pages/policy/output/Display.jsx
@@ -158,17 +158,24 @@ export function DisplayImpact(props) {
       <ErrorPage message="This service is temporarily unavailable. Please try again later." />
     );
   } else {
-    const { chart, csv } = getImpactReps(impactType, {
+    const value = getImpactReps(impactType, {
       impact: impact,
       metadata: metadata,
       policyLabel: policyLabel,
       mobile: mobile,
     });
-    pane = chart;
-    if (csv) {
-      downloadCsvFn = () => downloadCsv(csv(), filename);
+    if (value) {
+      const { chart, csv } = value;
+      pane = chart;
+      if (csv) {
+        downloadCsvFn = () => downloadCsv(csv(), filename);
+      }
+      downloadPngFn = () => downloadPng(filename);
+    } else {
+      pane = (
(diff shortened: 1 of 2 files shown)


src/pages/policy/output/Display.jsx
```diff
@@ -158,17 +158,24 @@ export function DisplayImpact(props) {
...
+    if (value) {
+      const { chart, csv } = value;
+      pane = chart;
+      if (csv) {
+        downloadCsvFn = () => downloadCsv(csv(), filename);
+      }
+      downloadPngFn = () => downloadPng(filename);
+    } else {
+      pane = (
+        <ErrorPage message="This service is temporarily unavailable. Please try again later." />
```
anth-volk, line 176: In order to match with our 404 page, since the pages this logic works for don't actually exist, could you change the message to:

Sorry, we can't find the page you're looking for

anth-volk (changes requested):
Thanks so much for this, @JiginJayaprakash! I just requested a minor change to do with wording. Otherwise, I think this would be good to go.

abhcs:
You may not even need to use the `map` in `ImpactTypes.jsx`. The exported `policyOutputs` map in `tree.jsx` may be a better choice for determining whether the URL is valid or not.

anth-volk:
Thanks for the insight, @abhcs. I'll take some time tomorrow and look into it and likely adjust the review accordingly, as cleaner code is definitely better.

anth-volk (changes requested):
Actually, to update my comments based on some insight from @abhcs - could you move the URL checking up into `PolicyOutput.jsx`? I think it would be far cleaner there. Thanks so much, @JiginJayaprakash.

PR #1196, opened 2024-01-16
1032 : fix for downloaded png

### LINKED ISSUE #1032: The description in the downloaded png for some charts goes out of frame
On saving the result for [this policy](https://policyengine.org/uk/policy?focus=policyOutput.cliffImpact&region=uk&timePeriod=2023&baseline=1&reform=42716) as a png, I get the following image:

cliffImpactPolicy #42716

Expected: The description should stay in the frame.

JiginJayaprakash:
Fixes #1032 

Fixes the png download issue

### DIFF
--- a/src/pages/policy/output/utils.js
+++ b/src/pages/policy/output/utils.js
@@ -14,6 +14,7 @@ async function downloadPng(filename) {
   if (downloadableContent) {
     const paddedDiv = document.createElement("div");
     paddedDiv.style.padding = "20px";
+    paddedDiv.style.width = "900px";
     paddedDiv.style.backgroundColor = "white";
     paddedDiv.style.display = "inline-block";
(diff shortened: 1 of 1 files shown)


MaxGhenis:
Thanks! Could you add some example results?

anth-volk (changes requested):
Unfortunately, this change is severely breaking; for a given policy, our charts previously looked like this:

genderDeepPovertyImpactPolicy #44170(3)
Not ideal.

Unfortunately, this edit causes the same chart to render like this:
genderDeepPovertyImpactPolicy #44170(5)
Less ideal.

I think the root of this problem actually lies in the component's entire structure, in that it takes a portion of the document that is nested in a complex manner and tries to merely format it as if it's the same portion of the page. I think there's probably a better way that is a bit more structured, and would love your contribution if you'd be willing to help build it.

PR #1225, opened 2024-01-19
Temporarily disable the download png button for policy output charts

### LINKED ISSUE #1222: Remove the "download png" button for policy output charts
Per https://github.com/PolicyEngine/policyengine-app/issues/966#issuecomment-1896643599, we should disable "download png" till we have a suitable fix for issues such as #966, #1032, #1205, #1206.

abhcs:
## Description

We fixed #1222 by not displaying the download png button for policy output charts.

## Screenshots

Normally the "download csv" button is the second button but now that we don't display the png button it is in first place:

Screenshot 2024-01-19 at 7 30 04 AM

## Tests

We tested that the button was not visible for 5 policy output charts.


anth-volk (approved):
Thanks for this, @abhcs! Looks good to go.

PR #1237, opened 2024-01-19
Add `embed=true` to API playground Streamlit URL

### LINKED ISSUE #1229: Add `?embed=true` to API playground Streamlit URL
The "API playground" at the bottom of [this page](https://policyengine.org/us/api) renders as merely an empty iframe

abhcs:
## Description

We fix #1229 by adding `?embed=true` to the API playground Streamlit URL.

## Screenshots

The playground is still visible after adding the parameter:

Screenshot 2024-01-19 at 12 51 55 PM

## Tests

It is hard to test this change -- it is hypothesized in https://github.com/PolicyEngine/policyengine-app/issues/1229#issuecomment-1900715529 that this will enable the user to see the hibernating Streamlit app and wake it up.  We will know whether this fix works or not once Streamlit sleeps again. If it does not work, the issue will need to be reopened.

### DIFF
--- a/src/redesign/components/APIDocumentationPage.jsx
+++ b/src/redesign/components/APIDocumentationPage.jsx
@@ -370,7 +370,7 @@ export default function APIDocumentationPage({ metadata }) {
       <Section title="API playground">
         <p>Try out the API in this interactive demo.</p>
         <iframe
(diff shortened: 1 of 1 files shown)


anth-volk (approved):
I certainly think this is worth merging, and if the embed fails down the road, we'll have to come back and reassess. Thanks @abhcs!

PR #1239, opened 2024-01-20
Display the reformed parameter properly in the `ParameterOverTime` component

### LINKED ISSUE #1238: Parameter history chart bug with multiple changes across periods
For example for the Wyden-Smith CTC reform: https://policyengine.org/us/policy?reform=44823&focus=gov.irs.credits.ctc.refundable.individual_max&region=us&timePeriod=2024&baseline=2

image

abhcs:
## Description

We fix #1238 by using an object with sorted keys for displaying reformed parameter values. This idea of using an object with sorted keys was already used for the baseline parameter but was not applied to the reformed parameter.

## Changes

- We add a new function `getSortedParameterValues` to `parameter.js`. This function returns a copy of `parameter.values` with sorted keys. This logic was already present in the `ParameterOverTime` component but we move it to `parameter.js` because it is a way to get a canonical representation of `parameter.values`.
- We use `getSortedParameterValues` to get a canonical representation of reformed parameter values instead of using the values directly in the chart.

## Screenshots

Screenshot 2024-01-19 at 9 53 47 PM

## Tests

We checked that the `ParameterOverTime` was displayed correctly for a couple of parameters.

### DIFF
--- a/src/api/parameters.js
+++ b/src/api/parameters.js
@@ -141,6 +141,25 @@ export function getReformedParameter(parameter, reforms) {
   return newParameter;
 }
 
+/**
+ *
+ * @param {object} parameter the parameter object
+ * @returns copy of parameter.values with sorted keys
+ *
+ */
+export function getSortedParameterValues(parameter) {
+  const values = parameter.values;
+  if (!values) {
+    return null;
+  }
+  return Object.keys(values)
+    .sort()
+    .reduce((obj, key) => {
+      obj[key] = values[key];
+      return obj;
+    }, {});
+}
+
 export function getNewPolicyId(countryId, newPolicyData, newPolicyLabel) {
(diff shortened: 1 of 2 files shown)


MaxGhenis:
Could we connect dec 31 to Jan 1 without the momentary gap going to the baseline?

abhcs:
The gap between dec 31 and jan 1 is due to the treatment of the intervals in `policy.reform.data` in a closed-open manner, i.e., the left end is inclusive but the right end is not. From the needs of this issue, I have gathered that it is better to treat the intervals as closed, i.e., both ends are inclusive. This yields the following graph:

Screenshot 2024-01-19 at 11 02 04 PM

MaxGhenis (approved):

PR #1249, opened 2024-01-22
Repair Reproduce in Python code

### LINKED ISSUE #653: Replace autogenerated Python code `calculate_dataframe` with `calculate`
MicroDataFrames are buggier than MicroSeries

### LINKED ISSUE #1179: Set `simulation.use_reported_state_income_tax = True` in `Reproduce in Python` for US nationwide reforms
`Reproduce in Python` does not currently reproduce the US microsimulation results in the web app, because we don't set the `simulation.use_reported_state_income_tax` parameter to `True`.

### LINKED ISSUE #1248: There's a strange trailing semicolon on the policy calculator's `Reproduce in Python` page
It's barely visible in the image below, a couple lines below the `COPY` button. I'm not entirely sure where it's from, but it's not just a hanging paragraph tag within the component.
Screen Shot 2024-01-22 at 3 17 54 PM

anth-volk:
## Description

Fixes #1179, fixes #653, fixes #1248. Previously, the code included in the `Reproduce in Python` page of the policy reform calculator displayed non-working code. 

## Changes

This PR repairs this code by ensuring that US contexts (those with region of "us" and "enhanced_us") have additional lines aimed at ensuring the baseline policy sets the parameter `simulation.use_reported_state_income_tax` to true. It also shifts from using `calculate_dataframe` to `calc` in order to prevent bugs associated with that operation, and removes a trailing semicolon from a different component. Finally, it ensures that the `year` property of the displayed code updates properly, and for policies with multiple parameters of different time periods, it ensures that `use_reported_state_income_tax` applies to the entire simulated period by iterating over the parameter list and setting its time period based upon that.

## Screenshots

A Loom video is available at https://www.loom.com/share/795b31c44954498f829a5690ebd53250?sid=a1d64b72-cdd0-49b6-9b66-b085ab5a5444.

## Tests

None have been included.

### DIFF
--- a/src/pages/policy/output/PolicyReproducibility.jsx
+++ b/src/pages/policy/output/PolicyReproducibility.jsx
@@ -106,6 +106,7 @@ function getBaselineDefinitionCode(region, policy) {
     "        value=True)",
     "    return parameters",
     "",
+    "",
     "class baseline_reform(Reform):",
     "    def apply(self):",
     "        self.modify_parameters(modify_baseline)",
--- a/src/pages/policy/output/PolicyReproducibility.jsx
+++ b/src/pages/policy/output/PolicyReproducibility.jsx
@@ -56,8 +56,7 @@ export default function PolicyReproducibility(props) {
           paddingTop: 30,
           marginBottom: 30,
         }}
-      >
-      </div>
+      ></div>
     </>
   );
 }
--- a/src/pages/policy/output/PolicyReproducibility.jsx
+++ b/src/pages/policy/output/PolicyReproducibility.jsx
@@ -4,6 +4,8 @@ import { defaultYear } from "data/constants";
 import { useSearchParams } from "react-router-dom";
 import colors from "../../../redesign/style/colors";
 
+const US_REGIONS = ["us", "enhanced_us"];
+
 export default function PolicyReproducibility(props) {
   const { policy, metadata } = props;
   const [searchParams] = useSearchParams();
@@ -72,7 +74,7 @@ function getHeaderLines(metadata) {
 }
 
 function getBaselineDefinitionCode(region, policy) {
-  if (region !== "us" && region !== "enhanced_us") {
+  if (!US_REGIONS.includes(region)) {
     return [];
   }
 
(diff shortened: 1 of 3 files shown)


src/pages/policy/output/PolicyReproducibility.jsx
```diff
@@ -56,10 +53,78 @@ export default function PolicyReproducibility(props) {
...
+      }
+    }
+  }
+
+  return [
+    `"""`,
+    "In US simulations, use reported state income tax liabilities",
+    `"""`,
+    "def modify_baseline(parameters):",
+    `    parameters.simulation.reported_state_income_tax.update(start=instant("${earliestStart}"), stop=instant("${latestEnd}"), value=True)`,
```
MaxGhenis, line 102: could we break these up so they fit on 80-character lines?

src/pages/policy/output/PolicyReproducibility.jsx
```diff
@@ -56,10 +53,78 @@ export default function PolicyReproducibility(props) {
...
+    "class baseline_reform(Reform):",
+    "    def apply(self):",
+    "        self.modify_parameters(modify_baseline)",
+    "",
+    "",
+  ];
+}
+
+function getImplementationCode(metadata, timePeriod) {
+  const isCountryUS =
```
MaxGhenis, line 114: also apply on line 76
```suggestion
  const US_REGIONS = ["us", "enhanced_us"];
  const isCountryUS = US_REGIONS.includes(metadata.countryId); 
```

MaxGhenis (changes requested):
Thanks - does this also affect the household side, with and/or without a reform?

anth-volk:
To answer your broader question, this does not affect the household side, as I was unaware of a need to change the code there, as well. Is that necessary?

MaxGhenis:
Oops you're right, household side is fine.

anth-volk:
I've spent quite a bit of time working on something way too nitpicky. At certain times, when editing a parameter with a name that's longer than 80 characters, the manual breaks embedded in this code don't actually limit the parameter name. I've written some code to fix this, but it's very finicky and I'm not in love with it. Can we push this, even though it doesn't fully accomplish the goal of splitting long parameter names into 80-char segments, instead of using the spaghetti I've written?

MaxGhenis:
Yes that was just for that one line if it was easy!

PR #1252, opened 2024-01-22
Improve parameter editor

### LINKED ISSUE #967: The parameter input box does not parse strings with commas or currency symbols

### LINKED ISSUE #1214: X'ing out the date in the policy parameter editor pages displays "invalid date" and "NaN" throughout the calendar component
This happens after you press the X button on the right side of the date input component:

Screen Shot 2024-01-18 at 11 29 57 PM

### LINKED ISSUE #1240: The parameter editing model is ill-defined in some cases
Any parameter is a step function from Date -> Value. A canonical representation of a parameter is provided below:

```
{
    "2013-01-01": 1000,
    "2018-01-01": 1400,
    "2022-01-01": 1500,
    "2023-01-01": 1600,
    "2024-01-01": 1700,
    "2026-01-01": 1000
}
```

The interpretation of the representation is that the parameter equals
- `1000` in the interval `["2013-01-01", "2018-01-01")`
- `1400` in the interval `["2018-01-01", "2022-01-01")`
- ...
- `1000` in the interval `["2026-01-01", inf)`
The parameter is undefined in `(-inf, "2013-01-01")`.

In the parameter editor, we allow the user to modify a parameter by specifying a new value for a chosen interval. For the parameter in our example, such an edit is ok if the interval intersects `["2013-01-01", inf)`. However, let's say that the user specifies the value `500` for `["2010-01-01", "2011-12-31"]`. Now the parameter must be defined for `["2010-01-01", "2011-12-31"]`, then undefined for `["2012-01-01", "2013-01-01")`, and then defined for `["2013-01-01", inf)`. This new parameter cannot be represented in the canonical form. Currently, we ignore the right endpoint of the user-specified interval, i.e., `"2011-12-31"`, and decide that the new parameter is given by:

```
{
    "2010-01-01": 500,
    "2013-01-01": 1000,
    "2018-01-01": 1400,
    "2022-01-01": 1500,
    "2023-01-01": 1600,
    "2024-01-01": 1700,
    "2026-01-01": 1000
}
```

There are two possible fixes for this error:
1. Enforce that each parameter is defined from `(-inf, inf)` at all times, perhaps by adding a mandatory entry of the form `"-inf": value` in each parameter.
2. Do not allow an edit to modify a region before the earliest point of definition, `"2013-01-01"` in the example.

### LINKED ISSUE #1241: Place `Current law` before `Reform` in policy parameter legend
Would be more natural imo

image

### LINKED ISSUE #1242: Use reform name in policy parameter chart
Instead of "Reform"

image

### LINKED ISSUE #1262: Change dashing and order of current law / reform parameter lines
| Attribute | Current | Requested |
| - | - | - |
| Current law line| Dashed | Solid |
| Reform line | Solid | Dashed |
| Top line | Current law | Reform |

image

### LINKED ISSUE #1275: The input field in the policy calculator is inaccessible
Take [this example](https://policyengine.org/us/policy?focus=gov.irs.income.bracket.rates.2). Type `13` in the field and press enter. The graph updates correctly, but the input field erroneously deletes the text to show a placeholder text `13%` instead. To edit the value to `13.5`, you must type `13.5` instead of appending `.5` to an existing `13`.

### LINKED ISSUE #1285: I can't move to a different parameter before entering the value for the current parameter
## Steps

1. Go to [this parameter](https://policyengine.org/us/policy?focus=gov.irs.income.bracket.thresholds.1.JOINT&region=us&timePeriod=2024&baseline=2).
2. Type a new value.
3. While the focus is on the input field, click on a different parameter in the left panel, e.g., `SEPARATE`.

The app momentarily goes to the page for the new parameter but returns immediately to the current parameter.

## Cause

Clicking on the parameter name in the left panel creates two events -- the blur event for the input field and the click event for the new parameter. The event handlers for these events try to change the URL but the blur event handler wins.

## Suggested fix

Don't change the URL on blur.

abhcs:
## Description

- We fix #1240 by
  - reducing the default range for the date picker in `ParameterEditor` to the current year
  - changing the allowed range for the date picker in `ParameterEditor` to `"2021-01-01" -> "2026-12-31"`.
- We fix #1241 by placing `Current law` before `Reform`
- We fix #1242 by changing `Reform` to `Policy #...`.
- We refactor the function `getReformedParameter` in `parameter.js` into an `IntervalMap` data structure. This leads to
  - simpler code for functions and components such as `getParameterAtInstant` and `ParameterEditor`, and
  - fewer errors because the data structure is tested thoroughly in `IntervalMap.test.js` (more tests should be added because `IntervalMap` is quite complex).
 - We remove `getSortedParameterValues` from `parameter.js` because `IntervalMap` provides sorted values.
 - We remove the `useEffect` block for setting dates in `ParameterEditor` because the heuristic of reading the first reform does not work: try changing a range such as `2027->2028` to a new value in [this example](https://policyengine.org/us/policy?reform=44823&focus=gov.irs.credits.ctc.refundable.individual_max&region=us&timePeriod=2024&baseline=2) and the dates erroneously jump back to `2023`.
 - We fix #1262 by changing the styles of the lines in the chart.
 - We fix #1285 by not changing the URL on blur.
 - We fix #967 and fix #1275 by using the `StableInputNumber` component instead of the `InputField` component.
 - We fix #1214 by using `defaultValue` in `RangePicker`.

## Screenshots

Screenshot 2024-01-22 at 1 58 53 PM

Note that this is the screenshot for the initial set of commits -- see discussion below for updated screenshots.

## Tests

Tests have been added for `IntervalMap`.
For the remaining changes, basic tests have been performed on the parameter editor but I am looking for feedback.

### DIFF
--- a/src/pages/policy/input/ParameterOverTime.jsx
+++ b/src/pages/policy/input/ParameterOverTime.jsx
@@ -1,9 +1,5 @@
 import Plot from "react-plotly.js";
 import { ChartLogo } from "../../../api/charts";
-import {
-  getSortedParameterValues,
-  getReformedParameter,
-} from "../../../api/parameters";
 import { getPlotlyAxisFormat } from "../../../api/variables";
 import useMobile from "../../../layout/Responsive";
 import useWindowHeight from "layout/WindowHeight";
@@ -12,7 +8,7 @@ import { plotLayoutFont } from "pages/policy/output/utils";
 import { localeCode } from "lang/format";
 
 export default function ParameterOverTime(props) {
-  const { parameter, policy, metadata } = props;
+  const { baseMap, reformMap, parameter, metadata } = props;
   const mobile = useMobile();
   const windowHeight = useWindowHeight();
 
@@ -23,23 +19,13 @@ export default function ParameterOverTime(props) {
     y.push(y[y.length - 1]);
   };
 
-  const values = getSortedParameterValues(parameter);
-  let x = Object.keys(values);
-  let y = Object.values(values);
+  const x = baseMap.keys();
+  const y = baseMap.values();
   extendForDisplay(x, y);
-  let reformedX;
-  let reformedY;
 
-  if (policy.reform.data[parameter.parameter]) {
-    const reformedParameter = getReformedParameter(
-      parameter,
-      policy.reform.data,
-    );
-    const reformedValues = getSortedParameterValues(reformedParameter);
-    reformedX = Object.keys(reformedValues);
-    reformedY = Object.values(reformedValues);
-    extendForDisplay(reformedX, reformedY);
-  }
+  const reformedX = reformMap ? reformMap.keys() : [];
+  const reformedY = reformMap ? reformMap.values() : [];
+  if (reformMap) extendForDisplay(reformedX, reformedY);
 
   let xaxisValues = reformedX ? x.concat(reformedX) : x;
   xaxisValues = xaxisValues.filter(
@@ -66,7 +52,7 @@ export default function ParameterOverTime(props) {
             },
             name: "Current law",
           },
-          policy.reform.data[parameter.parameter] && {
+          reformMap && {
             x: reformedX,
             y: reformedY.map((y) => +y),
             type: "line",
@@ -78,9 +64,7 @@ export default function ParameterOverTime(props) {
             },
             name: "Reform",
           },
-        ]
-          .reverse()
-          .filter((x) => x)}
+        ].filter((x) => x)}
         layout={{
           xaxis: { ...xaxisFormat },
           yaxis: { ...yaxisFormat },
--- a/src/pages/policy/input/ParameterOverTime.jsx
+++ b/src/pages/policy/input/ParameterOverTime.jsx
@@ -54,7 +54,6 @@ export default function ParameterOverTime(props) {
             type: "line",
             line: {
               shape: "hv",
-              dash: "dot",
             },
             marker: {
               color: style.colors.GRAY,
@@ -67,6 +66,7 @@ export default function ParameterOverTime(props) {
             type: "line",
             line: {
               shape: "hv",
+              dash: "dot",
             },
             marker: {
               color: style.colors.BLUE,
--- a/src/pages/policy/input/ParameterOverTime.jsx
+++ b/src/pages/policy/input/ParameterOverTime.jsx
@@ -5,10 +5,19 @@ import useMobile from "../../../layout/Responsive";
 import useWindowHeight from "layout/WindowHeight";
 import style from "../../../style";
 import { plotLayoutFont } from "pages/policy/output/utils";
-import { localeCode } from "lang/format";
+import { localeCode } from "lang/format"; /**
+ *
+ * @param {object} policy the policy object
+ * @returns the reform policy label
+ */
+function getReformPolicyLabel(policy) {
+  const urlParams = new URLSearchParams(window.location.search);
+  const reformPolicyId = urlParams.get("reform");
+  return reformPolicyId ? `Policy #${reformPolicyId}` : "reform";
+}
 
 export default function ParameterOverTime(props) {
-  const { baseMap, reformMap, parameter, metadata } = props;
+  const { baseMap, reformMap, parameter, policy, metadata } = props;
   const mobile = useMobile();
   const windowHeight = useWindowHeight();
 
@@ -62,7 +71,7 @@ export default function ParameterOverTime(props) {
             marker: {
               color: style.colors.BLUE,
             },
-            name: "Reform",
+            name: getReformPolicyLabel(policy),
           },
         ].filter((x) => x)}
         layout={{
--- a/src/pages/policy/input/ParameterOverTime.jsx
+++ b/src/pages/policy/input/ParameterOverTime.jsx
@@ -11,6 +11,7 @@ import { localeCode } from "lang/format"; /**
  * @returns the reform policy label
  */
 function getReformPolicyLabel(policy) {
+  if (policy.reform.label) return policy.reform.label;
   const urlParams = new URLSearchParams(window.location.search);
   const reformPolicyId = urlParams.get("reform");
   return reformPolicyId ? `Policy #${reformPolicyId}` : "reform";
--- a/src/__tests__/Bisection.test.js
+++ b/src/__tests__/Bisection.test.js
@@ -1,0 +1,49 @@
+const { bisect } = require("algorithms/Bisection");
+
+describe("bisection", () => {
+  test("small cases", () => {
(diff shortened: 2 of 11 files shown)


anth-volk:
I can go ahead and test this tomorrow, unless Max wants to look into this earlier

src/pages/policy/input/ParameterOverTime.jsx
```diff
@@ -1,18 +1,23 @@
...
-import { localeCode } from "lang/format";
+import { localeCode } from "lang/format"; /**
+ *
+ * @param {object} policy the policy object
+ * @returns the reform policy label
+ */
+function getReformPolicyLabel(policy) {
+  const urlParams = new URLSearchParams(window.location.search);
+  const reformPolicyId = urlParams.get("reform");
+  return reformPolicyId ? `Policy #${reformPolicyId}` : "reform";
```
anth-volk, line 16: Instead of this, could you display the full name here if there is one, followed by the policy number if there isn't a name? And perhaps there's a good way to truncate this if the name is too long?

anth-volk (changes requested):
Thank you for this, @abhcs, I think this is a substantial improvement over the current logic that we have, and I think it's a great choice to couple all of the internal logic and data representation together into this new data structure. 

One thing I want to address - I think the inclusive end date logic, even though it's the way to go, introduces an internal contradiction. If you take [this](https://policyengine.org/us/policy?reform=44823&focus=gov.irs.credits.ctc.refundable.individual_max&region=us&timePeriod=2024&baseline=2) test link and run it with your changes, when I edit the date of a parameter to run until Dec. 31, the old parameter is displayed as running from Dec. 31 of Year X to Jan. 1 of year X+1, which I can change manually. However, the right-hand sidebar never recognizes the Dec. 31-Jan. 1 artifact, instead indicating that the policy correctly ends Dec. 31. Is there a way to handle this issue? 

This further manifests itself when creating new policies. For example, let's say I raise IRS income tax bracket 2 from 12% to 25% for the 2024 fiscal year (Jan. 1-Dec. 31), then raise it to 28% for the 2025 fiscal year (Jan. 1-Dec. 31). The graph then shows that the policy has not been raised for Dec. 31, 2024, to Jan. 1, 2025. However, the right sidebar will in fact show that the policy changes only run to Dec. 30.

I'd be curious if you have some proposals as to how to handle these issues. More broadly, though, I think the fundamental problem here is that we're trying to treat yearly variables as if they're daily variables. I opened a new issue (#1261) to rectify this, as I think we should instead let the user choose full years, then we'll handle the Dec. 31/Jan. 1 discrepancy ourselves behind-the-scenes.

I also left one minor comment regarding policy naming, and I'm waiting on a response from Max regarding whether or not we're looking for the `reform` plot's line to overlap over current policy, or merely for the `reform` label to come first in the key, though I imagine the line should be on top, as well.

Thank you very much for this, it really is a great PR, I just want to iron out a few things before proceeding.

MaxGhenis:
As some context on parameter periods: in the US, tax parameters align with the calendar year, but many benefit parameters change mid-calendar-year (e.g. USDA updates SNAP parameters on October 1). In the UK, both tax and benefit parameters change mid-year, e.g. tax parameters change each April.

Typically these changes take effect for a full year after their date, so we may be able to simplify the UX by making that a default, where the user selects a beginning period and we fill in the end date with that plus a year (minus one day). Setting an as-of date and leaving the parameter at that value in perpetuity from there is probably the most common use case, so we could also hide the end date selector by default. I'm not sure the best design but hope this background helps.

anth-volk:
Perhaps if we even did one of the following:

* Allowed users to select a series of months, starting from the relevant start date
* Allowed users to select a start date, followed by a duration in years and/or months (we could adapt this depending on the parameter); the user could then click a `+` icon to add another parameter change

abhcs:
The odd artifact on Dec 31 was due to a simple bug in `onChange`. Please let me know if you see any other odd behavior or if I missed something in your previous post.

New screenshot:
Screenshot 2024-01-23 at 12 34 28 PM

abhcs:
On a side note, the right-hand sidebar has at least two problems.

Screenshot 2024-01-23 at 12 42 19 PM

1. It is inefficient: `Child tax credit refundable maximum amount` is repeated for each interval
2. It is inaccurate: the `from` part of the message only considers the original value in the first modified interval.

Instead of the current behavior, it might be better to 
- show a  clickable list of modified parameters in the right sidebar
- show the modified intervals in the main window with a +/- icon to add/delete changes

anth-volk:
In response to your comments on the carousel, I will say that converting to a list is definitely a common piece of commentary, but I'd want @MaxGhenis's input before opening an issue, though I'm all for it

MaxGhenis:
We'd like to replace the carousel with a full-page description of the policy reform. I think we might overflow in the right sidebar too, especially given the other info there and with mobile in mind.

abhcs:
I claim that all the relevant information for the policy editing is present in the following (rough) sketch:

Untitled

I have removed one panel. The names of the charts can be shown *after* we have computed the impact.

MaxGhenis:
A full-page view defining the reform would give the room for each of these, for each parameter:
1. Full path (`Amount` isn't informative)
2. Current and reform values, in each period if it changes multiple times

abhcs:
I agree that changes such as the ones discussed in the last six comments need a lot of planning, but let's get back to the contents of this PR.

anth-volk:
@abhcs Would you mind adding that you fix #1262 in your description at the top of this PR? Thanks again for including it.

abhcs:
The dot is expected in this case -- when I zoom in I get this chart, which seems accurate:

Screenshot 2024-01-24 at 7 51 52 AM

abhcs:
The last commit is a potential fix for #967 and #1275:

Screenshot 2024-01-24 at 10 56 11 AM

MaxGhenis:
Filed https://github.com/PolicyEngine/policyengine-app/issues/1282 for a future full-page reform description

abhcs:
My initial implementation of IntervalMap was based on a sorted array but I was using the inbuilt `findIndex` and `findLastIndex` functions, which ignore order in the arrays. In the last commit, I replaced these functions with bisection.

anth-volk:
Will review later today

anth-volk:
Just reviewed. This does fix #1277.

anth-volk (approved):
This looks great, @abhcs! Thanks for all of your revisions and for fixing a series of important issues the app faced.

PR #1256, opened 2024-01-23
removed copy button and updated the note

### LINKED ISSUE #1254: Copy Button still exists At the Bottom of the reproduce python code block
https://github.com/PolicyEngine/policyengine-app/assets/46224156/8cb2642a-d4eb-4b5b-9d1f-626ff6e57e95

I have recently worked on adding a copy button on the top right corner of the codeblock and its working for both household and policy and after that I removed the copy button from `PolicyReproducibility.jsx` file. But this only removed the button from `policy` page and not in the household page.

dotslashbit:
## Description

fixes #1254 

## Screenshots

Screenshot 2024-01-23 at 11 26 08 AM


dotslashbit:
@MaxGhenis Is this what you want me to do?

src/pages/household/output/HouseholdReproducibility.jsx
```diff
@@ -81,7 +80,7 @@ export default function HouseholdReproducibility(props) {
       </Helmet>
       <ResultsPanel
         title="Reproduce these results"
-        description="Run the code below into a Python notebook to reproduce the microsimulation results."
+        description="Run the code below to a Python notebook to reproduce the results."
```
anth-volk, line 83: I believe Max meant to say "in" here, not "to," so would you mind changing this once more?

anth-volk (changes requested):
Thanks for this @dotslashbit! It looks great, but just one thing (other than the changing "to" to "in") - the screenshot you provided shows the code preview existing in a scroll container, that itself sits inside a scroll container. Could you look into this, and if the second container is introduced by your changes, could you remove it? If not, we can just open a new issue for that one, and honestly I'd love your contributions to remove it, if you're willing.

src/pages/household/output/HouseholdReproducibility.jsx
dotslashbit: The two scrolls were already there, at some point I thought I would ask you about this, but then forgot about it? Should I open a new issue and work on it?

anth-volk: Please, if you'd be willing

anth-volk (changes requested):
I'm sorry, some code I pushed yesterday that's currently in your fork's `master` branch was breaking and was reverted very late last night. Could you rebase off of that, then re-request review? Unfortunately it breaks exactly what has to be checked.

anth-volk:
I'll be sure to confirm that this works one last time later today, then we should be good to go! Thanks @dotslashbit!

anth-volk (approved):
Thanks so much for your contribution to this, @dotslashbit! All looks good, approved.

PR #1257, opened 2024-01-23
Simplify variable editor logic

### LINKED ISSUE #1132: `addVariable` does not add the default state CA to the household
Issue #1057 was closed through PR #1061 by adding using `variable.defaultValue` in the `HouseholdVariableEntityInput` component. However, this fix seems redundant because `addVariable` is responsible for adding variables (including `state_name`) with default values to all entities. `addVariable` ignores the default value for `state_name` because on [line 353](https://github.com/PolicyEngine/policyengine-app/blob/d6181d2676f995fe6e9c20d2c6a83195afd2a4e2/src/pages/household/input/VariableEditor.jsx#L353) it finds that `state_name` is *not* an input variable and uses the default value null.

Why is `state_name` not an input variable? Is the notion of input and non-input variables important? The fix for #1057 should be revisited with these questions in mind.

abhcs:
## Description

We fix #1132 by simplifying the logic in the `addVariable` function and the `HouseholdVariableEntityInput` component. As mentioned in the description for #1132, our fix for #1057 violated the Single Responsibility Principle. The current fix addresses #1057 at a deeper level.

## Changes

The cause of #1132 is the failure of `addVariable` to carry out its responsibility of adding variables to entities with good default values. `addVariable` ignores the default value for the `state_name` variable because it is not an "input variable". `state_name` is not an input variable because it has a formula:

https://github.com/PolicyEngine/policyengine-us/blob/e73ce75edac5e7bb88b9ea538a012cae83825f9e/policyengine_us/variables/input/geography.py#L19

and any variable with a formula is not an input variable according to core logic. However, this definition of input variables is irrelevant to the front end. For the front end, if we are showing the variable in the input portion of the household calculator, then it is an input variable. Therefore, we remove the unnecessary check in `addVariable` and the unnecessary branch in `HouseholdVariableEntityInput`.

## Screenshots

Screenshot 2024-01-23 at 7 58 22 AM

## Tests

We tested the variable editor a few times with different number of household members.

### DIFF
--- a/src/pages/household/input/VariableEditor.jsx
+++ b/src/pages/household/input/VariableEditor.jsx
@@ -234,8 +234,6 @@ function HouseholdVariableEntityInput(props) {
     defaultValue = inputValue;
   } else if (simulatedValue !== null) {
     defaultValue = simulatedValue;
-  } else if (variable.defaultValue !== undefined) {
-    defaultValue = variable.defaultValue;
   }
 
   if (defaultValue === null) {
@@ -367,20 +365,9 @@ export function addVariable(householdInput, variable, entityPlural, year) {
       possibleEntities.forEach((entity) => {
         // If the variable isn't already stored in the situation...
         if (!(variable.name in householdInput[entityPlural][entity])) {
-          // If the basic input is an "input variable" (input by user)...
-          if (variable.isInputVariable) {
-            // Then add it to the relevant part of the situation, along with
-            // its default value
-            newHouseholdInput[entityPlural][entity][variable.name] = {
-              [year]: variable.defaultValue,
-            };
-          } else {
-            // Otherwise, add it to the relevant part of the situation, along with
-            // a null value
-            newHouseholdInput[entityPlural][entity][variable.name] = {
-              [year]: null,
-            };
-          }
+          newHouseholdInput[entityPlural][entity][variable.name] = {
+            [year]: variable.defaultValue,
+          };
         }
       });
     }


anth-volk:
Before reviewing this, I'm going to request @nikhilwoodruff's sign-off on removing the check on whether a variable is an input variable. I think you're exactly right on doing so, but just want a double-check from someone with more familiarity.

src/pages/household/input/VariableEditor.jsx
```diff
@@ -367,20 +365,9 @@ export function addVariable(householdInput, variable, entityPlural, year) {
       possibleEntities.forEach((entity) => {
         // If the variable isn't already stored in the situation...
         if (!(variable.name in householdInput[entityPlural][entity])) {
```
anth-volk, line 367:
```suggestion
        if (!(variable.name in householdInput?.[entityPlural]?.[entity])) {
```

Just in case there are more profound errors passed by some other part of the app, I'd suggest this to prevent runtime app failure

anth-volk (changes requested):
Thank you for this, @abhcs, now that Nikhil has confirmed that treating input variables and non-input variables the same way shouldn't create issues, I'm able to review. I did request one minor change to introduce optional chaining to accommodate for objects that are malformed or formed in unexpected ways, but otherwise this looks great, thanks so much.

src/pages/household/input/VariableEditor.jsx
abhcs: This optional chaining is unnecessary due to line 363 and where `VaraibleEditor` is used in `HouseholdPage`.

I understand the need for caution, but using optional chaining everywhere is not the answer. I once worked on a legacy codebase where every line was surrounded by 
```
if (x != null) ...
```
and this style was enforced during code reviews. Mandates like this lead to ugly conditional code, IMO. The preconditions, postconditions, and invariants of functions/components should be used to avoid errors including accessing empty objects. Of course, this process is error-prone -- that is why we need good reviews and tests.

Even if we were to use optional chaining here, which is unnecessary, we would need
```
if (!(variable.name in (householdInput?.[entityPlural]?.[entity] || {}))) {
```
instead of the suggested change.

anth-volk: Fair enough, and admittedly, the portion of the code this would touch is mostly controlled by the back end, anyway. I'll merge as-is.

anth-volk (approved):
Thanks again for this, @abhcs, it's very helpful.

PR #1263, opened 2024-01-23
changed link color to blue with an underline

### LINKED ISSUE #1051: Links need a uniform style that has different font color and underlining
For example, on the interstitial for the policy calculator:

Screen Shot 2023-12-28 at 12 33 23 PM

`Learn more` is a link, but it's not clear that it is one, hindering accessibility. Ideally, these should be underlined and take a different color (most ideal is system's default blue, in my opinion). This isn't the only example of this, and we should standardize these across the app. One potential option would be to emulate what the blog posts do.

dotslashbit:
## Description

fixes #1051 

## Screenshots
Screenshot 2024-01-24 at 2 09 34 AM


dotslashbit:
@anth-volk I noticed that there are two styles directory, one is inside redesign and another is outside redesign and the style for `a` is coming from redesign's styles directory and not from the outside one.
Do we even need that other styles directory?

anth-volk:
@dotslashbit We had kept both because the older, un-redesigned components used references to the older. That said, if you'd be willing to open an issue and convert all the old references to new ones, I know it's a low-level task, but it's something I've been trying to promote because having two is confusing.

dotslashbit:
@anth-volk I can work on it, but sorry if this is a silly question, but can you explain what do you mean by "un-redesigned components used references to the older".
I think it would be better if you could open an issue on this, and then I'll work on it

anth-volk:
@dotslashbit When we implemented a redesign back in the fall, we only redesigned a portion of the site. We left the calculator components intact, and those parts are what I mean by "un-redesigned." No worries at all for asking, I'm using a made-up word, after all, and sure, I'll open the issue.

anth-volk:
Oh, and I'll likely review this early tomorrow

anth-volk (changes requested):
Thanks for this contribution, @dotslashbit, it's definitely an accessibility improvement! One thing I will say is that by coloring all `a` tags, this unfortunately adds coloring and underlines to anything that includes a link, even if it's a button (for example, if you run your test code, you'll notice that the "Support Transparent Policy Access" button on the Donate page also receives this styling). This is of course an issue in almost all sites.

Something I've done before to fix this is to put in place the rules for `a` tags, as you've done, then override the rules for `a`s inside `button` tags. However, since Ant Design uses `div` elements for its buttons, could you do the reverse and only apply the styling to `a` tags inside of `p` tags, and then I can look again and see if any don't get applied properly? It's not a great solution, but it's something I hadn't thought of when we discussed this to begin with.

Additionally, would you mind adding, then applying, an inverted style that keeps the words their defined font color? This would be for, for example, the `hello@policyengine.org` under `How to Donate` on the same Donation page.

Thanks so much, and I wish I had thought of these tricky cases before you started, as I would've mentioned them.

anth-volk:
To be clear, I meant, could you find all places where the background is dark and add a custom CSS ruleset, something like "link-inverted," and then apply it to them? If you'd be willing

dotslashbit:
ok, will do that

anth-volk:
Thanks @dotslashbit! I should have some time later today to review this.

anth-volk (changes requested):
I'm sorry, @dotslashbit, this unfortunately lacks the handling we had discussed for links on dark backgrounds. Would you be able to add that before I push this? It could take the form of a separate CSS class added to the relevant links that merely overrides the color definition provided for `p a` links

dotslashbit:
Actually, I haven't worked on it yet, I'll do it by tomorrow, I should have convert it to draft.

dotslashbit:
@anth-volk I have updated the code, please check

anth-volk (approved):
Thanks so much for this, @dotslashbit, it looks great! By the way, have you already joined our Slack? If not, feel free to let me know and we'll figure out a way to get you a link to join, if you'd like to.

dotslashbit:
@anth-volk I would love to join, please send me the link, thanks

PR #1319, opened 2024-02-03
Disable zoom on y-axis for policy output charts

### LINKED ISSUE #1310: The y-axis ticks in policy output charts don't respond to plotly zoom events
## Steps

1. Open a policy output chart, e.g., [this one](https://policyengine.org/uk/policy?focus=policyOutput.wealthDecileAverageImpact&reform=39411&region=uk&timePeriod=2024&baseline=1).
2. Zoom in on a range on the y-axis, say `[7.5, 9.5]`.

## Output

The y-axis tick numbers do not respond to the zoom level, and consecutive tick values are often duplicates:

Screenshot 2024-02-02 at 8 46 35 AM

## Expected

Duplicate tick values are not helpful. However, maintaining tick values for different zoom levels is work. I don't think that this work is justified because zooming in on a y-axis range in a bar chart is not a useful functionality.

## Suggestion

We should disable zooms for the y-axis for bar charts using `yaxis: {fixedrange: true}`.

abhcs:
## Description

We fix #1310 by disabling zoom on the y-axis. The x-axis can still be zoomed:

Screen Recording 2024-02-02 at 6 29 06 PM

### DIFF
--- a/src/pages/policy/output/CliffImpact.jsx
+++ b/src/pages/policy/output/CliffImpact.jsx
@@ -96,6 +96,7 @@ function ImpactPlot(props) {
         yaxis: {
           title: "Relative change",
           tickformat: `+,.${ytickPrecision}%`,
+          fixedrange: true,
         },
         ...(useHoverCard
           ? {}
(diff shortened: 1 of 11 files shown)


MaxGhenis (approved):
Thanks, this is a frequent issue on mobile. I'll try it out more and file an issue if we should do the same for the x axis.

PR #1324, opened 2024-02-05
Button Component Refactored

### LINKED ISSUE #950: Button components should be rewritten against Ant Design
This has implications for form submission actions, event firing, and accessibility

dotslashbit:
## Description

fixes #950 

## Changes

Please describe the changes in detail.


dotslashbit:
@anth-volk Points 2,3,4 are more or less done, I have added this button on the about page just to check how it's looking and its hover effect and if I'm getting the context menu and everything seems to be working fine. Let me know your thoughts on this.

I'll work on the 1st point that you mentioned from now on

dotslashbit:
@anth-volk I have now worked on all the 4 points, please check and let me know if I'm missing something

dotslashbit:
@anth-volk I have removed the NewButton component and replaced all the code from this new button to the old one. The Button component is the base button and all the derived button components are working fine.

Please take a look and let me know if there's any issue that I missed to address. Thanks

src/controls/Button.jsx
```diff
@@ -35,60 +36,53 @@ const buttonStyles = {
...
+        (e.currentTarget.style.borderColor =
+          buttonStyles[type].standardBackgroundColor)
+      }
       size={size ? size : width ? `${width}px` : "300px"}
       onClick={onClick}
-      aria-label={text}
     >
-      {text}
-    </HoverBox>
+      {text ? text : `🎉New Button`}
```
anth-volk, line 83:
```suggestion
      {text}
```
In this case, I actually think we're fine with the app breaking, since a button with placeholder text is as bad for the UX as a broken button, in my opinion

src/controls/Button.jsx
```diff
@@ -35,60 +36,53 @@ const buttonStyles = {
...
-      }
-      hoverStart={hoverStart || "left"}
+    <AntButton
       style={{
-        alignItems: "center",
         display: "flex",
-        backgroundColor:
-          backgroundColor || buttonStyles[type].standardBackgroundColor,
-        color: "white",
+        borderRadius: "0",
```
anth-volk, line 51:
```suggestion
        borderRadius: "0",
        border: "none",
```
To remove the automatic 1px white border that gets added

src/controls/Button.jsx
```diff
@@ -35,60 +36,53 @@ const buttonStyles = {
...
+        backgroundColor: colors.TEAL_ACCENT,
```
anth-volk, line 52:
```suggestion
        backgroundColor: buttonStyles[type].standardBackgroundColor,
```
This is to return the ability to set buttons of different types, like "disabled" or "secondary"

src/controls/Button.jsx
```diff
@@ -35,60 +36,53 @@ const buttonStyles = {
...
-    width,
-    type,
-    size,
-    height,
-    hoverStart,
-    backgroundColor,
-    activeBackgroundColor,
-    style,
-  } = props;
+const Button = (props) => {
```
anth-volk, line 39:
```suggestion
export default function Button(props) {
```
`Button` is a full-fledged React 16.8+ hook-based component, so this declaration style better matches industry standards

anth-volk (changes requested):
Thanks for your contribution on this, @dotslashbit, it's something we've been wanting to do, so I'm glad we're doing it. In addition to the inline comments, I'm hoping you can do the following:

* Change the button in the navbar to also function the same way as these buttons; the navbar button must be somehow different, but it would be poor form to merge this code without editing that
* Allow button text to wrap to prevent the following issue at the bottom:
Screen Shot 2024-02-09 at 6 56 05 PM

The second point is part of a bigger issue we have where we place quite a bit of text on each button and should probably use less. Something we should probably do in the future is to write something to downsize text to a certain point, then start line-breaking.

dotslashbit:
@anth-volk I have updated all the nav buttons for both the desktop and mobile views, please take a look and let me know if there's any changes needed to be done

anth-volk (changes requested):
I'm sorry, when I said this earlier:

`Change the button in the navbar to also function the same way as these buttons; the navbar button must be somehow different, but it would be poor form to merge this code without editing that`

I just meant that we currently use the `HoverBox` component only for that purpose, and that it might be better to move it to using the new `Button` component. In truth, we're looking for that menu (probably the left-hand menu, as well) to look exactly the same as it did before. If you're unable to maintain the look and feel, could you revert to what was there before? Sorry for the confusion, and thanks for these contributions.

dotslashbit:
@anth-volk I have reverted the code

anth-volk:
Don't worry, I am still reviewing this - there's just one minor oddity that I want to iron out (I think it's Ant Design's fault), as buttons are a major part of the UX, so I want to make sure we transition smoothly. I'll probably revisit it Monday and hopefully get this merged then.

src/controls/Button.jsx
```diff
@@ -36,59 +37,52 @@ const buttonStyles = {
...
   }
 
   return (
-    <HoverBox
-      hoverBackgroundColor={
-        activeBackgroundColor || buttonStyles[type].hoverBackgroundColor
-      }
-      hoverStart={hoverStart || "left"}
+    <AntButton
       style={{
```
anth-volk, line 49:
```suggestion
      style={{
        alignItems: "center"
```

(Unless there's an issue somewhere else in the code with doing this - I haven't found one) Could you add back in the `alignItems` setting? The arrows on the arrow buttons float upward too much if it's removed.

anth-volk (changes requested):
Okay, there's one minor change that we need, then I think this is good to go:

In the ArrowButton component, if you could set "alignItems" to "center", this will prevent the arrow from basically floating a bit upward

Thanks for bearing with me, Sahil, it's much appreciated.

dotslashbit:
@anth-volk I have updated the code as required

anth-volk (approved):
Thank you very much for this, @dotslashbit, it's a big help!

PR #1339, opened 2024-02-09
Fix formatting for EITC policy parameters

### LINKED ISSUE #1336: Buggy formatting for two EITC policy parameters
Three EITC parameters display buggily:
* [Maximum investment income](https://policyengine.org/us/policy?reform=2&focus=gov.irs.credits.eitc.phase_out.max_investment_income)
* [Phase-out start joint filer bonus (no children)](https://policyengine.org/us/policy?reform=2&focus=gov.irs.credits.eitc.phase_out.joint_bonus%5B0%5D.amount)
* [Phase-out start joint filer bonus (with children)](https://policyengine.org/us/policy?reform=2&focus=gov.irs.credits.eitc.phase_out.joint_bonus%5B1%5D.amount)

Example:
image

Here are the issues:
1. Number entry box is divided by $1,000, or $10,000 for maximum investment income, and rounded
2. Time period goes from 1969 to 2099 instead of normal 2012-2030
3. A reform line is shown even though no reform is provided (matching baseline)

I don't see anything unusual in the parameter files themselves that would cause this issue:
* https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/parameters/gov/irs/credits/eitc/phase_out/max_investment_income.yaml
* https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/parameters/gov/irs/credits/eitc/phase_out/joint_bonus.yaml

I also haven't seen it in other parameters yet, having spot-checked a few.

abhcs:
## Description

We fix #1336. The root cause is a bug in the `IntervalMap` constructor. The bug is pretty bad, so it is a surprise that the editor was working for some parameters!

## Changes

- create a deduplicated copy of the input array in the `IntervalMap` constructor.
- use bisection in `IntervalMap.get` for efficiency.
- add more tests for `IntervalMap`.

## Screenshots

Screenshot 2024-02-08 at 8 05 16 PM

## Tests

The parameters in #1336 were tested.

### DIFF
--- a/src/__tests__/IntervalMap.test.js
+++ b/src/__tests__/IntervalMap.test.js
@@ -62,6 +62,22 @@ describe("IntervalMap construction", () => {
       [10, "f"],
     ]);
   });
+
+  test("unsorted to sorted", () => {
+    const m = new IntervalMap(
+      [
+        [2, "b"],
+        [1, "a"],
+        [3, "c"],
+      ],
+      (x, y) => x - y,
+    );
+    expect(m.toArray()).toStrictEqual([
+      [1, "a"],
+      [2, "b"],
+      [3, "c"],
+    ]);
+  });
 });
 
 describe("IntervalMap operations", () => {
--- a/src/__tests__/IntervalMap.test.js
+++ b/src/__tests__/IntervalMap.test.js
@@ -33,6 +33,35 @@ describe("IntervalMap construction", () => {
       [2, "b"],
     ]);
   });
+
+  test("ten entries, one duplicate", () => {
+    const m = new IntervalMap(
+      [
+        [1, "a"],
+        [2, "b"],
+        [3, "c"],
+        [4, "d"],
+        [5, "e"],
+        [6, "f"],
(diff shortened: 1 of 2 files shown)


MaxGhenis:
Thanks @abhcs! It seems like it's working for parameters with only more recent values provided, but I've since found it's affecting a fair bit more of them so I'm also surprised we didn't catch it sooner.

abhcs:
I understand now why it was working for some parameters and not for others. The bug manifests itself only on parameter functions with at least `10` steps. [This parameter](https://policyengine.org/us/policy?focus=gov.irs.income.bracket.thresholds.2.JOINT) had fewer than `10` steps and was working ok. The parameters in the issue are more jagged (`>= 10` steps) -- so the bug shows up there.

I believe that the bug would have been caught earlier if we were using TypeScript or if we added more test cases.

MaxGhenis (approved):
Thanks I'm going to merge given the severity of the bug but welcome further review later from others

PR #1341, opened 2024-02-10
Fix bugs in Parameter Editor

### LINKED ISSUE #1340: Some edits don't work in parameter editor
## Steps

In [this example](https://policyengine.org/us/policy?reform=2&focus=gov.irs.credits.eitc.phase_out.max_investment_income),

- change the start and end dates to `2023-01-01` and `2028-12-31` respectively, and 
- change the value to `$11,000`.

No change is observed in the graph and the edit is not registered.

## Expected

The function should change in the given range.

## Cause

A bug in the difference calculation between baseline and reform policies.

abhcs:
## Description

We fix #1340.

## Changes

- We rewrite `IntervalMap.js` and `Bisection.js` in TypeScript, and compile them to JavaScript with strict type-checking enabled. The use of TypeScript is motivated by bug #1336, which was fixed in PR #1339 in part by using the correct `for` loop construct for arrays. One can verify that using the incorrect `for` construct in the current PR results in a compilation error -- thus many errors are ruled out due to type checking.
- We rewrite the test files for `IntervalMap.js` and `Bisection.js` in TypeScript. This change revealed that we were using the `test.each` construct incorrectly in these files -- we fix this error.
- We add a `minus` function to `IntervalMap`, which computes the difference between two maps more reliably than the function `diff` in `ParameterEditor`.
- We add tests for `minus`.
- We use the improved `IntervalMap` with the `minus` function to fix #1340.

## Screenshots

Screen Recording 2024-02-09 at 6 34 33 PM


anth-volk:
I'm someone who vastly prefers working with strongly typed languages - if I had my preference, we'd rip out much of the app and API and replace them with TypeScript and something static, respectively - but I have to question the benefit here in rewriting just two files in TypeScript, especially for an issue originally driven not by type mismatch issues or a series of complex interconnected classes and/or interfaces, where typing can be very difficult to manage without these sorts of constraints, but instead by syntax issues. I acknowledge that said syntax issues would fail on compile, but is it worth introducing the added complexity of TypeScript without utilizing it anywhere else in the bundle? Most importantly, though, the vast majority of contributors come to PolicyEngine from one of two spaces: the data science realm (i.e., they're much more familiar with Python, anyway, and don't always write idiosyncratic JavaScript, let alone know the quirks of TS) or the beginner space (so they often just don't know much of anything). 

Again, don't get me wrong - I would much prefer static typing myself, and I see a lot of value in TS overall. I just want to identify these negatives at this early stage and see if they're worth the positives. I'm also going to rope in @nikhilwoodruff on this, as I'm curious if he has an opinion.

abhcs:
I am not using the class feature of ts but I am using [generics](https://www.typescriptlang.org/docs/handbook/2/generics.html). Note that generics are also used in the C++ standard lib, especially for algorithm implementation. Generics and type-checking increase my confidence in the correctness of this code that is difficult to get right, as evidenced by the number of bugs encountered on this topic. The ts features are dev dependencies only -- see `package.json`; so, it should not affect the bundle. The way I am using ts does not force anyone to use ts -- it just enables optional usage in critical sections. Even in the algorithms folder, one can stick to js if that is their preference.

I agree with some of the qualms of @anth-volk in using ts but I view the use here as driven by needs.

abhcs:
Improved test coverage in the last commit:

Screenshot 2024-02-11 at 12 38 14 PM

MaxGhenis:
Actually I'm going to merge it given the severity of the bug. Anthony please file an issue to take a different approach if you'd like to suggest that.

MaxGhenis (approved):

anth-volk:
Yeah, I just wanted to ensure higher-level visibility and comfort with the addition of TypeScript, particularly from Nikhil, but as long as we have top-level support, I'm fine with it

PR #1342, opened 2024-02-11
Increase Jest timeout parameter

### LINKED ISSUE #1218: Increase Jest timeout parameter
It seems counterintuitive to time out a test out at only a few seconds that, for example, includes back-end fetches, as quite a few do. It seems more logical to raise that number to, say 10s?

ken1009us:
## Description

Fixes #1218

Increase Jest timeout parameter

## Changes

Add the timeout param

## Tests

Pass all tests

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -8,6 +8,7 @@ const esModules = [
 module.exports = {
   modulePaths: ["<rootDir>/src"],
   testEnvironment: "jsdom",
+  testTimeout: 10000,
   transform: {
     "^.+\\.[jt]sx?$": "babel-jest",
   },


anth-volk:
Thanks so much for your first contribution, @ken1009us! Unfortunately, as I'm looking through this, I'm not sure it's an issue anymore, as we've moved to just manually running `jest` within the repo instead of using `react-scripts`. I hate to do this, but I'm going to close this PR and issue; that said, there are many more issues you might contribute to, and we'd really appreciate your help!

anth-volk:
Apologies, I take that back, we actually ran into this. It would appear Jest just sets the timeout by default, which is why I wasn't finding a line of code creating it. I'll probably review this tomorrow.

PR #1368, opened 2024-02-15
Emit ESNext code in TypeScript compilation

### LINKED ISSUE #1352: `Bisection.test.js` uses `require`
Should be import, as the package is ES6-compliant

abhcs:
## Description

We fix #1352 by emitting ESNext code on compiling TypeScript. The new module options in `tsconfig.json` are chosen by consulting the following links:

https://stackoverflow.com/questions/71463698/why-we-need-nodenext-typescript-compiler-option-when-we-have-esnext
https://www.typescriptlang.org/docs/handbook/modules/theory.html

## Changes

- The module options are changed.
- The ts files in `src` dir are recompiled.

## Tests

`npm run test` succeeds.

### DIFF
--- a/src/__tests__/Bisection.test.js
+++ b/src/__tests__/Bisection.test.js
@@ -1,24 +1,23 @@
-Object.defineProperty(exports, "__esModule", { value: true });
-var Bisection_1 = require("algorithms/Bisection");
+import { bisect } from "algorithms/Bisection";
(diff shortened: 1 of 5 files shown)


anth-volk (approved):
This all makes a lot of sense. Thanks for the contributions @abhcs!

PR #1377, opened 2024-02-20
Improve PolicyNamer

### LINKED ISSUE #1165: Every time a user creates a new policy name, at least 2 policies are created
In the policy calculator, if you try creating a unique policy, then changing its name, it won't be readily apparent, but at least two policies will be created and sent to the back end. For example, I created US policy 44150 with the name "Test Policy" and an increased IRS tax bracket 4 percentage value of 49.2%. Policies 44151, 44152, 44153, and 44154 all contain the exact same information. This is likely driven by some sort of poorly synced effect on the front end that's repeatedly creating all of these.

### LINKED ISSUE #1166: When a user edits policy parameters then names the policy, multiple policies are created
This is related to #1165 and #972. Per @patrickhanl, it would appear that whenever a user edits policy parameters for a policy they've created, then names the policy, multiple corresponding back-end policies are created. Ideally this should not occur.

### LINKED ISSUE #1190: Create some sort of "rename" button that becomes "submit"

### LINKED ISSUE #1292: In `InputField`, enter events sometimes fire twice
In the `InputField` component, an `Enter` event fires twice, with my guess as to why being that they also emit a `blur` event when the input item becomes unfocused. This is most important for the `PolicyNamer` component on the policy editor's right-hand sidebar, where it calls two POST requests to create or update a household.

anth-volk:
## Description

Fixes #1165
Fixes #1166
Fixes #1190
Fixes #1292 

## Changes

Previously, when a user would attempt to rename policies, various issues would arise: numerous duplicates would be created; if a policy already had been renamed, it couldn't successfully be renamed again; and it was unclear for users that there even was a method to rename. 

While some of these issues were diminished in previous API improvements, many continued to emerge from the `InputField` component, which had various limitations. This PR replaces `InputField` (which is currently only utilized in the `PolicyNamer` component on the policy calculator pages) with the `InputText` component, aimed purely at textual input, thereby limiting its scope from the prior `InputField`. The rewritten component does the following:

* Replaces the previous custom-built input field with a standardized Ant Design version
* Builds in an optional right-hand submit button, which the `PolicyNamer` component is rewritten to utilize
* Replaces `PolicyNamer`'s use of `onChange` props with `onPressEnter` and `onClick` events so as to make clear when submit events are firing
* Removes any use of `blur` events, as these are fired very frequently and are not the clearest input method for the user
* Adds validation to `PolicyNamer`, which is currently only used to prevent renaming a policy to an empty string, but could be employed in the future to deal with #972 
* Removes the title from the rename box and places it above to make it clear to the user what the purpose of the `PolicyNamer` component is
* Builds in optional disabling of the submit button when the input field is empty, which the `PolicyNamer` component is written to utilize
* Adds margin to the "search for a policy" box to align it with all other elements

## Screenshots

A video of the component in action is available below. This video also demonstrates that only one renamed policy is created.
https://github.com/PolicyEngine/policyengine-app/assets/14987227/d4a12f8d-4db1-418d-841f-ca49f6bb71a9

## Tests

The existing test for `InputField` has been renamed to `InputText` and updated to match the new desired behavior.

### DIFF
--- a/src/__tests__/InputText.test.js
+++ b/src/__tests__/InputText.test.js
@@ -46,6 +46,6 @@ describe("Should take inputs", () => {
     expect(testProps.onPressEnter).not.toHaveBeenCalled();
 
     fireEvent.keyDown(input, { key: "Enter" });
-    expect(testProps.onPressEnter).toHaveBeenCalled();
+    expect(testProps.onPressEnter).toHaveBeenCalledTimes(1);
   });
 });
--- a/src/__tests__/InputText.test.js
+++ b/src/__tests__/InputText.test.js
@@ -1,5 +1,5 @@
 import { render, screen, fireEvent } from "@testing-library/react";
-import InputField from "../controls/InputField";
+import InputText from "../controls/InputText";
 import { useSearchParams } from "react-router-dom";
 
 jest.mock("react-router-dom", () => {
@@ -24,7 +24,7 @@ describe("Should take inputs", () => {
     };
 
     setup = (props) => {
-      const utils = render(<InputField {...props} />);
+      const utils = render(<InputText {...props} />);
       const input = screen.getByPlaceholderText(testProps.placeholder);
 
       return {
--- a/src/__tests__/InputText.test.js
+++ b/src/__tests__/InputText.test.js
@@ -10,7 +10,6 @@ jest.mock("react-router-dom", () => {
     useSearchParams: jest.fn(),
   };
 });
-const patternProp = { pattern: "%" };
 const testInput = "Test Input";
 
 describe("Should take inputs", () => {
@@ -20,7 +19,7 @@ describe("Should take inputs", () => {
   beforeEach(() => {
     testProps = {
       placeholder: "Test Placeholder",
-      onChange: jest.fn(),
+      onPressEnter: jest.fn(),
     };
 
     setup = (props) => {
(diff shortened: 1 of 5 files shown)


MaxGhenis (approved):
Thanks Anthony - let's merge this given the severity of the bugs and also continue to think about how to iterate on the UX. For instance I think we can refine this to avoid adding a third place in the right sidebar where a policy name shows up.

Please go ahead and merge it if you're comfortable with the PR, e.g. the lack of testing (I know this component is untested today so seems OK for now).

anth-volk:
With this PR, I actually removed the way the renaming box featured the name by default, meaning that the right-hand bar only displays the policy name once (twice if the policy search bar is opened, but then I think that should also be redone to not contain the policy name as a placeholder). 

And with regard to the testing, moving forward, I will aim to extend coverage when rewriting and/or building. While I tested the underlying logic within `InputText`, I hadn't considered testing the logic that handles shading on hover, which is the part that remains coverage-less.

PR #1385, opened 2024-02-21
refactored desktop calculator button

### LINKED ISSUE #1379: COMPUTE POLICY IMPACTS button in navbar is not `Button` and thus doesn't match other buttons
The "COMPUTE POLICY IMPACTS" button in the navbar is not actually a `Button` component, meaning that changes to the `Button` component itself do not affect this item. This is more visible than previously as a result of #1324, which redoes buttons to work more cohesively. This is also true of the button's smaller form, a small calculator icon on mobile.

The issue wasn't addressed in #1324 itself, as I thought it prudent to merge the changes, then add this item as a separate issue, so that we could move the ball forward on the important changes included in #1324.

dotslashbit:
## Description

refactored the compute policy navbar button 

fixes #1379 

## Screenshots

Screenshot 2024-02-21 at 10 03 57 PM

### DIFF
--- a/src/controls/LinkButton.jsx
+++ b/src/controls/LinkButton.jsx
@@ -28,6 +28,7 @@ export default function LinkButton(props) {
     backgroundColor,
     activeBackgroundColor,
     link,
+    style,
   } = props;
 
   const navigate = useNavigate();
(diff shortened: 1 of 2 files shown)


dotslashbit:
@anth-volk I haven't modified the mobileCalculator component as that's just a link with an image

anth-volk (approved):
This is perfect, @dotslashbit! Thank you very much.

PR #1414, opened 2024-02-23
Properly handle policy parameters with numbers in Reproduce in Python

### LINKED ISSUE #1391: Reorder tests to match the folder structure of the repo
At the moment, the tests are just basically strewn about in the __tests__ folder

### LINKED ISSUE #1392: Ensure that parameters with numbers in them are properly rendered for the `Reproduce in Python` output pages for policy reforms
Let's say a user raises the US income tax bracket 2 rate to 15%. The corresponding `Reproduce in Python` output would be the following:

```
from policyengine_us import Microsimulation
from policyengine_core.reforms import Reform
from policyengine_core.periods import instant


"""
In US nationwide simulations,
use reported state income tax liabilities
"""
def modify_baseline(parameters):
    parameters.simulation.reported_state_income_tax.update(
        start=instant("2024-01-01"), stop=instant("2024-12-31"),
        value=True)
    return parameters


class baseline_reform(Reform):
    def apply(self):
        self.modify_parameters(modify_baseline)


def modify_parameters(parameters):
    parameters.simulation.reported_state_income_tax.update(
        start=instant("2024-01-01"), stop=instant("2024-12-31"),
        value=True)
    parameters.gov.irs.income.bracket.rates.2.update(
        start=instant("2024-01-01"), stop=instant("2024-12-31"),
        value=0.15)
    return parameters


class reform(Reform):
    def apply(self):
        self.modify_parameters(modify_parameters)
baseline = Microsimulation(reform=baseline_reform)
reformed = Microsimulation(reform=reform)
baseline_person = baseline.calc("household_net_income",
    period=2024, map_to="person")
reformed_person = reformed.calc("household_net_income",
    period=2024, map_to="person")
difference_person = reformed_person - baseline_person
```

However, if you run this in a Colab notebook, you'll receive a syntax error at this line:

```
parameters.gov.irs.income.bracket.rates.2.update(
```
because of the digit in the parameter. 

The code that generates the `Reproduce in Python` block should handle this and output the following:
```
parameters.gov.irs.income.bracket.rates["2"].update(
```
per Nikhil

anth-volk:
## Description

Fixes #1392 
Fixes #1391
The root of this issue lies in `policyengine-core` and the country packages, and I think that further discussion is warranted to determine if we shouldn't write a check within the `ParameterNode` class initializer in `core` to prevent this. 

At present, there are a few parameters (including the quite useful IRS tax parameters in the US package) that utilize a raw integer as a name. While Python allows users to name an attribute as such, it is not possible to access the resulting attribute using dot notation, as all of the following methods of accessing it fail:

* `parameter.1.something`
* `parameter[1].something`
* `parameter(1).something`
* `parameter."1".something`
* `parameter["1"].something`
* `parameter[(1)].something`

This PR attempts to solve this issue for users by combining `reduce` from the `functools` package and `getattr` to access the numeric parameter. This fix should be short-term, as addressing the underlying problem in the data packages would be a more effective solution, but this PR ensures that any users (especially those unfamiliar with Python syntax) are still able to easily reproduce code output in a Colab notebook, etc.

## Changes

* Adds a function to determine if a parameter name contains a number
* Changes `getReproducibilityCodeBlock` to add a line to import `reduce` from `functools` if a parameter contains a number
* Changes the same function to rewrite the reform code block's parameter accessing line to use `reduce` instead of the standard method if a parameter name contains a number

## Screenshots

A video comparing the Reproduce in Python pages for params containing numbers and those not containing them is available below
https://github.com/PolicyEngine/policyengine-app/assets/14987227/a1dd2a80-639a-4995-b1cc-ba7f9ffd3a16

## Tests

Various tests are added for `getReproducibilityCodeBlock`

### DIFF
--- a/src/__tests__/__setup__/sampleData.js
+++ b/src/__tests__/__setup__/sampleData.js
@@ -9,7 +9,7 @@ export const baselinePolicyUK = {
     label: "Current law",
     id: 1,
   },
-}
+};
 
 export const baselinePolicyUS = {
   baseline: {
@@ -34,8 +34,8 @@ export const reformPolicyUK = {
     data: {
       "sample.reform.item": {
         "2020.01.01": true,
-        "2022.01.01": true
-      }
+        "2022.01.01": true,
+      },
     },
     label: "Sample reform",
     id: 0,
@@ -52,8 +52,8 @@ export const reformPolicyUS = {
     data: {
       "sample.reform.item": {
         "2020.01.01": false,
-        "2022.01.01": false
-      }
+        "2022.01.01": false,
+      },
     },
     label: "Sample reform",
     id: 0,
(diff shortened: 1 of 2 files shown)


nikhilwoodruff:
Thanks but just checking that `parameter.children["1"]` doesn't work?

anth-volk:
The Colab notebook at https://colab.research.google.com/drive/103q2I7Gdjmt3L5eYa0TB2857jrTzNqve?usp=sharing demonstrates what happens if you try to access the parameter that way. The notebook demonstrates that you'll receive a TypeError, as the code is attempting to access a characteristic within an instance of a class, as opposed to an element of a dictionary.

anth-volk:
Ah, I see, I misunderstood you there. Let me switch the PR over to draft (I may also just trash it) and start over on this.

anth-volk:
I'm going to dump this PR and open a new one, separating out the changes for each of the two issues

Issue #1426, opened 2024-02-28
Policy renaming function should hide the input bar until a user selects an edit icon


anth-volk:
On the right-hand panel, screenshotted below:
<img width="349" alt="Screen Shot 2024-02-28 at 10 09 20 PM" src="https://github.com/PolicyEngine/policyengine-app/assets/14987227/a771d4c6-82bf-46ef-99f4-e71833c2e453">
We should probably put an icon of a pencil either next to the name or all the way to the right of it, and when a user presses it, it should allow renaming, then there should be a way of closing the renaming panel, as well. Additionally, on hover, it should display a tooltip saying "rename policy". This would remove some of the clutter and complex functionality of the right-hand bar.

dotslashbit:
@anth-volk working on it

MaxGhenis:
Thank you for filing this issue. Closing as part of the `policyengine-app` (v1) wind-down: policyengine.org is now served by [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2), and this repository is being archived, so this issue is frozen against a codebase that no longer deploys anywhere.

If this is still relevant on the current site, please re-file it against [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2) — we'd welcome it there. Apologies that this sat open through the migration.

PR #1431, opened 2024-02-29
added a close option

### LINKED ISSUE #1428: Allow "find an existing policy" box to be closed
At the moment, if the user opens this box on the right-hand sidebar, there's no option for closing it

dotslashbit:
## Description

fixes #1428 

## Screenshots

https://github.com/PolicyEngine/policyengine-app/assets/46224156/463ccddf-7e3d-4427-880e-9d2b1bf8e37c

### DIFF
--- a/src/pages/policy/PolicyRightSidebar.jsx
+++ b/src/pages/policy/PolicyRightSidebar.jsx
@@ -568,6 +568,14 @@ export default function PolicyRightSidebar(props) {
           </a>
         </div>
       )}
+      {showReformSearch && (
+        <div style={{ textAlign: "center", width: "100%" }}>
+          {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+          <a href="#" onClick={() => setShowReformSearch(false)}>
+            Close
+          </a>
+        </div>
+      )}
       <PolicyDisplay
         policy={policy}
         metadata={metadata}


anth-volk:
@dotslashbit Before reviewing this - does the right-hand panel always have the strange white border around the reforms for you? And if so, which browser are you on?

dotslashbit:
@anth-volk yes, it still shows that white border even though I'm executing localhost from the master branch. I'm using Firefox on MacBook Air m2

anth-volk:
Does it occur for you when loading the live app, as well, or only when running locally?

anth-volk:
Just figured it out: do you have an option enabled to always show scrollbars on your MacBook? If so, I think what's happening is that these have `overflow` set to `scroll`, as opposed to `overflow-y`. I'll submit a new issue on this.

Otherwise, I'm going to wait a few hours to review this, as I want to pitch a different approach to fixing the right-hand sidebar this that's a bit less piecemeal. If you'd be willing, I'm thinking something like what you've done for #1426, but for this, as well, but I need to pitch the idea first.

dotslashbit:
Its not in always mode
Screenshot 2024-03-18 at 21 39 14

anth-volk:
That is puzzling. Even so, I was able to reproduce by:
1. Setting my scrollbars to always on (that allowed me to see them)
2. Removing the relevant scroll options, which removed the unnecessary bars entirely

anth-volk:
Unfortunately, I'm going to close this (and the original issue) in light of #1485. Thanks for your work on this, Sahil.

PR #1432, opened 2024-02-29
Fix: Added Icon and Tooltip

### LINKED ISSUE #1426: Policy renaming function should hide the input bar until a user selects an edit icon

dotslashbit:
## Description

fixes #1426 

Added an edit icon with tooltip and when the user submits by hitting enter it updates the name accordingly and goes back to the default look

## Screenshots

https://github.com/PolicyEngine/policyengine-app/assets/46224156/a3048abf-39f8-4d5a-bc05-a2db2c958027


MaxGhenis:
Could you align the text with the left of the text box?

anth-volk:
@dotslashbit Based on the screenshot above, I think you might want to rebase and see how the renaming component has changed.

Additionally, I didn't provide too much detail in the original issue as I had been thinking of working on it while doing another bug. It's completely fine that you want to - it's just that I had envisioned this working a bit differently. Ideally, I think we should keep the renaming component the same as it looks currently in the app, but hide it until the user clicks the pencil icon. The pencil should then become an X, allowing the user to close the editing pane.

Also, could you make the pencil more of a dark gray?

dotslashbit:
@anth-volk ok got it, will give you an update by tomorrow, and if you want to work on this, then let me know I'll close this.

Waiting for your response.

anth-volk:
@dotslashbit Happy for you to take it on, it's great to have your contributions to it, just wanted to let you know why the technical definition for it was so lacking.

dotslashbit:
@anth-volk I have updated the code, please take a look and let me know

https://github.com/PolicyEngine/policyengine-app/assets/46224156/0d2ecbcf-438b-4804-9928-16704c0771d9

anth-volk (changes requested):
Thanks so much for this, @dotslashbit. Could you do just a couple minor things?

1. Could you drop the color on the pencil and X to something like 80% gray? Just so the title still stands out.
2. Could you have the pencil sit against the text, as opposed to all the way to the right?
3. Could you have the box automatically close when the user presses to submit a new name?

Otherwise, this looks great.

MaxGhenis:
Sorry I'm popping in late here, but why do we have a separate text box for editing, rather than editing the name directly (as is the standard for renaming interfaces)? Is it a technical issue?

anth-volk:
Doing it this way has a few ease-of-use advantages for users who are new to the app, in my opinion. 

First, while a user is renaming the policy, it's abundantly clear what the state of the policy's name is, and if a mistake is made, if they hit the Escape key, or something else, it's clear if the change to the name has been applied or not. 

Because of this clarity, when we provide custom error handling, it's also clear if a change has been applied or not. Currently, we only test for whether or not a policy's name is empty, but when we start testing for duplicates, this will allow a user to know that their name has not been changed.

Additionally, though less important, this allows a user to more easily copy and paste the older name while updating it.

anth-volk:
To add to this, it mirrors a very common pattern used by Slack, Google, etc., whereby instead of opening a renaming box in place of the current text, an alert is thrown up. This just implements that pattern in the component, instead of in a popup.

MaxGhenis:
I agree that a popup is another common UX for renaming things. Can we adopt either that or the inline edit? I'd like to only adopt standard UX so the user has familiarity, and I've never seen this approach. (Let me know if this is modeled after another product though.)

anth-volk:
Frankly, I can't think of any other product with this method. In that case, @dotslashbit, would you be willing to do the following:

* Instead of displaying the box below, replace the existing text with a full-width box, placing the existing title as a placeholder
* Add a highlight over the X button that says "Cancel"

If not, too, I'm happy to take it on, I know this has had quite a few stylistic changes

anth-volk:
Closing, as this is incredibly deprecated unfortunately

PR #1451, opened 2024-03-05
Remove duplicate policies search

### LINKED ISSUE #1425: Prevent searching of completely duplicate policies
This is related to #972. The app should filter the policy results that are shown to only show policies that are not the same in every way besides their ID (including policy label). A series of duplicate policies were created as the result of a couple of bugs that have since been cleared up. However, it would be best not to remove duplicate policies from the database so that users can still access said policies by URL, but we do want to prevent the confusion of having a series of completely duplicate policies being shown in the search box for existing policies.

anth-volk:
## Description

Fixes #1425. 

## Changes

At the moment, the front-end application only receives the name and policy ID for each policy from the API's search endpoint, making it impossible to determine which policies are truly duplicates and which merely have the same name. As such, changes were made to the API that allow a query parameter to be passed to provide only "unique" policies, i.e., those with a distinct combination of label and policy (disregarding ID), to be returned. These changes were pushed in [#1268](https://github.com/PolicyEngine/policyengine-api/pull/1268).

This PR adds the query param to enable this unique policy filtering.

## Screenshots

A video of the code in action, followed by the previous behavior, followed by the code in action again is available below.
https://github.com/PolicyEngine/policyengine-app/assets/14987227/c31054eb-7d43-4952-ae4e-4d27d35879fe

## Tests

Because the unique filtering is done on the back end, tests were created in [#1268](https://github.com/PolicyEngine/policyengine-api/pull/1268) to ensure that when the uniqueness filtering param is provided, the API does, in fact, return unique policies. 

I also spent a few hours trying to create a front-end test in Jest that ensured that the PolicySearch component actually searches when a user types in a value and displays the proper list of output values, but Ant Design's `Autocomplete` component is incredibly opaque, making it very difficult to test whether or not the output value list is actually displayed. As such, no new front-end tests are being added with this PR.


anth-volk:
Merging per conversation with Nikhil

PR #1465, opened 2024-03-07
Fix API page

### LINKED ISSUE #1412: API returns different result than the one displayed on the API page
For the example input, the `snap` computed by the API is `2833.5`. On running the same example locally, I get

```js
{
    "message": null,
    "result": {
        "people": {
            "child": {
                "age": {
                    "2024": 5
                }
            },
            "parent": {
                "age": {
                    "2024": 30
                },
                "employment_income": {
                    "2024": 20000
                }
            }
        },
        "spm_units": {
            "spm_unit": {
                "members": [
                    "parent",
                    "child"
                ],
                "snap": {
                    "2024": 2334.0
                }
            }
        }
    },
    "status": "ok"
}
```

## Expected

The outputs should be consistent.

### LINKED ISSUE #1423: Sample API documentation code in `Calculate household-level policy outcomes` for the UK isn't valid within the UK API
This is because the sample household is hard-coded and does not change depending on country context

### LINKED ISSUE #1434: Restyle the JSON blocks on the API page
Currently, they don't really align with our design, but they also don't align with the various (also mismatching) API explorer pages at the end of the page

### LINKED ISSUE #1460: Variable cards on `APIDocumentationPage` don't properly display `period` value
Ran into this while building tests for it. Every card on the `APIDocumentationPage` displays the `period` value of each variable card as `None`.

### LINKED ISSUE #1461: Cards on `APIDocumentationPage` without given unit display empty space for unit
This should probably be something like "N/A" instead

anth-volk:
## Description

Fixes #1460 
Fixes #1434 
Fixes #1423 
Fixes #1412 
Fixes #1461 

## Changes

This PR redoes the sample API data segment of the API page by rewriting the component to instead fetch a sample output when the user loads the page. It also restyles the JSON blocks to make the page more mobile-friendly and to accord better with the PolicyEngine style, fixes the UK's sample data to be valid for the UK, provides a brief line of description for the sample fetch, and fixes a bug on the page's variable cards.

## Screenshots

The new page in action is available in the video below.

https://github.com/PolicyEngine/policyengine-app/assets/14987227/92a0fef5-ad9d-4af0-ad16-f971547b63fa

## Tests

Includes tests for the more important functionality of the `APIDocumentionPage` component, raising the page's coverage from 0% to around 55%, as measured by covered lines.


MaxGhenis:
Could you look for more idiomatic UI components for expanding boxes like this, rather than our CTA button?

anth-volk:
I'm glad this came up, as I think it'll help us refine our brand, as well. I'll come back later today and share some screenshots of possible changes, but in the mean time, we have three other components that use something similar: 

Here, on the reproduce pages
Screen Shot 2024-03-08 at 3 17 53 PM

Here, on the calculator itself
Screen Shot 2024-03-08 at 3 18 02 PM

And on the right policy sidebar
Screen Shot 2024-03-08 at 3 18 09 PM

Do you feel the buttons in any of these positions should be changed to also distinguish them from the CTA? Could definitely see shifting some to a darker blue variety instead

anth-volk:
One option: use more of our secondary styling options, though I think there could be value in changing the coloring. Here's one look:
Screen Shot 2024-03-08 at 4 42 32 PM

And another:
Screen Shot 2024-03-08 at 4 41 06 PM

There's probably also value in having a kind of "tertiary" look for options on smaller dialog boxes and the like, which I think would work well here. Here's that look, and I think this is by far the best of the three:
Screen Shot 2024-03-08 at 4 56 58 PM

MaxGhenis:
I agree the last one is best, but still think we could be more idiomatic.

If it's an accordion for the code block, this is the relevant component: https://ant.design/components/collapse

If it instead expands the partially-expanded code block, I think you'd use https://ant.design/components/typography#ellipsis - below is some sample code i asked chatgpt to spin up

and for copying code, i like how chatgpt does it, putting `Copy code` in a header ribbon
image

Code for expandable partly-visible code block:

```jsx
import React, { useState } from 'react';
import { Typography, Button } from 'antd';

const { Paragraph } = Typography;

const ExpandableCodeBlock = ({ code }) => {
  const [expanded, setExpanded] = useState(false);

  const toggleExpand = () => {
    setExpanded(!expanded);
  };

  return (
    <>
      <Paragraph
        ellipsis={expanded ? false : { rows: 2, expandable: false, symbol: 'more' }}
        code={true}
        style={{ fontFamily: 'monospace' }}
      >
        {code}
      </Paragraph>
      <Button type="link" onClick={toggleExpand}>
        {expanded ? 'Show Less' : 'Show More'}
      </Button>
    </>
  );
};

// Example usage
const exampleCode = `
def calculate_sum(a, b):
    return a + b

# Example usage
result = calculate_sum(5, 7)
print("The sum is:", result)
`;

const App = () => (
  <div style={{ margin: 20 }}>
    <ExpandableCodeBlock code={exampleCode} />
  </div>
);

export default App;
```

anth-volk:
Ah, I now better understand what you're getting at. I had previously looked into `Accordion`, but there's no option to partially expand it, but I will look into `Typography.ellipsis`.

anth-volk:
As I'm playing with this, part of the issue here, too, is that I think we want the default behavior to be a limited scroll box. `Typography.Paragraph`'s ellipsis feature merely shows the top text, preventing scrolling. My thinking is that the default scroll allows easier interaction for mobile users (where these boxes will be stacked) and allows for desktop users to scroll through and copy text if they want, especially the specific text of parameters, if desired, without being forced to expand the whole block.

anth-volk:
I've now modified this to look like the image below:

Screen Shot 2024-03-11 at 10 55 56 PM

Unfortunately, I was never able to find a way to make `Typography`'s ellipsis feature work, as it requires the code be fed in as a string and any code syntax highlighting package that was applied would prevent this. Ant Design also has no code block formatting, only inline, and if merely inputting the stringified JSON into any of `Typography`'s elements, the element would remove and/or ignore the stringify function's whitespace additions, making the code unreadable.

However, this replacement component I've added sets us up well for converting this section into an editable in-page API playground that would replace the Streamlit, removing issues we've had with the Streamlit not waking up or its iframe not rendering properly and improving the user experience in general.

nikhilwoodruff:
Low priority but can we use the existing PE colours (or just not purple)?

anth-volk:
@nikhilwoodruff What do you think of this theme instead? https://thememirror.net/espresso This is probably the bluest one I've found that doesn't require implementing our own. That said, if preferred, it can be done, it will just take more time.

anth-volk:
This now has the following theme:

Screen Shot 2024-03-18 at 11 40 34 PM
While this does contain a bit of red, this theme is probably the closest to our own without us having to write our own, and it does allow the data to pop a bit.

nikhilwoodruff (approved):

Issue #821, opened 2023-10-27
Tables are currently broken


nikhilwoodruff:
Tables don't show up. The cause is a bug in `remark-gfm` ([issue](https://github.com/remarkjs/react-markdown/issues/771))

PR #1507, opened 2024-03-26
Migrate to react-markdown 9.0.0

### LINKED ISSUE #1478: Migrate react-markdown to v9
This will require a few changes, as delineated in dependabot's description at https://github.com/PolicyEngine/policyengine-app/pull/1436

dotslashbit:
## Description

fixes #1478

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "react-ga": "^3.3.1",
         "react-helmet": "^6.1.0",
         "react-hook-reading-time": "^1.0.0",
-        "react-markdown": "^8.0.7",
+        "react-markdown": "^9.0.0",
         "react-plotly.js": "^2.6.0",
         "react-responsive-carousel": "^3.2.23",
         "react-router-dom": "^6.22.2",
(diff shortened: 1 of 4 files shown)


anth-volk:
I'm going to convert this to draft, as my understanding is that it's not quite ready yet, but feel free to convert back if I'm wrong

anth-volk:
Found the issue on this: in #821, we pinned the version we use for `remark-gfm` to <4. To be compatible with `react-markdown`, it should actually be pinned to `^4.0.0`, and when I looked to see if the original issue still occurs, it appears that it does not. As such, could you just raise the version of `remark-gfm`, then confirm that fixes it, then convert this PR to "ready"?

dotslashbit:
@anth-volk I have updated the code as required and confirmed that the error is not there anymore, thanks

anth-volk:
Perfect. Would you mind just resolving the conflict here, I'm guessing by installing locally and/or force-pushing on this branch? It may also require rebasing off master. After that, should be good to merge, I think. Thanks for your contributions, @dotslashbit.

anth-volk:
Thanks for this, @dotslashbit. Unfortunately, the tests fail because of an inherent issue within Jest - it still, still, unbelievably doesn't support ES6 imports, meaning we have to utilize Babel to transform certain code into ES6 import modules. However, we also have to manually specify that Babel not transform certain modules that are already only ES6-compliant, having abandoned the old CommonJS "require" structure. Unfortunately, it looks like this version of `react-markdown` is one such package, and this will require adding not only it, but also some packages it relies upon, to an ignorelist.

If you'd like to do this, I'm happy to tell you how to. However, it's quite mundane, so I'm also happy to do so, though probably not until tomorrow. Up to you.

dotslashbit:
@anth-volk Please do it as it'll be faster, thanks

anth-volk:
@dotslashbit Just finished. Merging.

anth-volk (approved):

dotslashbit:
@anth-volk Thank you so much

PR #1526, opened 2024-04-01
Make default start date and end date reflect previously input value

### LINKED ISSUE #1520: The default start and end date in the policy parameter editor don't reflect previously input value
I believe this may be a regression. Currently, if a user clicks on a preexisting parameter, instead of the default start and end dates being populated from that parameter, they default to `2024-01-01` and `2024-12-31`, respectively. An example of that behavior is shown in the image below. Ideally, they should be populated based on that parameter.

Screen Shot 2024-03-29 at 12 30 45 AM

cats256:
## Description

Closes #1520. This is not a complete solution yet. This is for reference in #1520 to discuss the complete solution. As of the initial PR, this correctly makes the start date and end date reflect the previously input value. However, it does not update the input number correctly when clicking a preexisting parameter and then another preexisting parameter, since the way the app passes in the input number is through StableInputNumber defaultValue prop, instead of the value prop, which causes the value state variable to be unchanged even when what's passed into defaultValue prop changed. The issue described can be seen more clearly in the video below.

## Changes

I create two new search parameters, startDate and endDate of a single policy to allow ParameterEditor access to the previous input value or single policy start and end date. I moved setting searchParams in Carousel.Item to setting searchParams in SinglePolicyChange (SinglePolicyChange is a grandchild component of Carousel.Item) to accommodate setting two new parameters, startDate and endDate. I, then, update ParameterEditor accordingly so that when a user clicks on a preexisting parameter, it uses startDate and endDate search params so startDate and endDate reflect previously input values.

## Screenshots

https://github.com/PolicyEngine/policyengine-app/assets/59489624/7d991972-6a41-44df-8f33-455126e44a6f

## Tests

Manual testing and with make test (all tests passed).

### DIFF
--- a/src/pages/policy/PolicyRightSidebar.jsx
+++ b/src/pages/policy/PolicyRightSidebar.jsx
@@ -265,8 +265,17 @@ function PolicyNamer(props) {
 }
 
 function SinglePolicyChange(props) {
-  const { startDateStr, endDateStr, parameterMetadata, value, paramLabel, parameterName, reformId, region, timePeriodYear } =
-    props;
+  const {
+    startDateStr,
+    endDateStr,
+    parameterMetadata,
+    value,
+    paramLabel,
+    parameterName,
+    reformId,
+    region,
+    timePeriodYear,
+  } = props;
   const [searchParams, setSearchParams] = useSearchParams();
   const oldVal = getParameterAtInstant(parameterMetadata, startDateStr);
   const oldValStr = formatVariableValue(parameterMetadata, oldVal);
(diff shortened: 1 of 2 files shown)


anth-volk:
Delaying merge based on discussion in #1520. I'm going to set the status of this to "Delayed merging," but not set this to draft, since it is a completed PR, just need to see if we should collectively take a more comprehensive approach to #1520 and some other interconnected issues, as well.

anth-volk:
Closing, as this would require fixing merge conflicts, particularly in the parameter editor itself

PR #1530, opened 2024-04-01
Display label metadata rather than filename for scale parameters

### LINKED ISSUE #1475: Display `label` metadata rather than filename for scale parameters
For example, this parameter has the label `EITC maximum amount by number of children`

https://github.com/PolicyEngine/policyengine-us/blob/90965abc185bcb9eb866d78c7e613d5d7bd6cdd6/policyengine_us/parameters/gov/irs/credits/eitc/max.yaml#L97

But we display just `Max` from `max.yaml`:

image

https://policyengine.org/us/policy?reform=2&focus=gov.irs.credits.eitc

abhcs:
## Description

We fix #1475 by using the correct label for scale parameters when constructing the parameter tree.

## Screenshots

Screenshot 2024-04-01 at 1 10 00 PM

### DIFF
--- a/src/api/parameters.js
+++ b/src/api/parameters.js
@@ -29,7 +29,9 @@ export function buildParameterTree(parameters) {
     for (const key of pathComponents.slice(0, -1)) {
       cumulativePath += key;
       const fixedCumulativePath = cumulativePath;
-      let label = key;
+      let label =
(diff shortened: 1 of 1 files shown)


anth-volk (approved):
What an improvement this is for UX. Thanks so much, @abhcs, this is much appreciated.

PR #1544, opened 2024-04-03
Improve ParameterEditor date selection

### LINKED ISSUE #1261: Feature request: yearly parameters should have yearly inputs instead of single date-driven ones
As I use the parameter editor, I'm struck by how many of the parameters we're editing take place over an entire fiscal year, but require the user to input data for individual dates. Furthermore, the individual date input is incredibly finicky due to the inherent challenge of what to do with the interval between Dec. 31st of given year X and Jan. 1st of given year X+1. Moving forward, I think it'd be great to have an input where the user chooses a year or series of years for these yearly parameters to occur over, removing the Dec. 31st/Jan. 1st guesswork, removing possible misinterpretation of the data (what if I decide to input May 1st, even though that's not a "valid" input from a legislative standpoint?), and just making it far easier to select years (no need to click 3-4 times to select start and end dates). Furthermore, we could limit the selectable years to the years the variable is calculated over.

anth-volk:
## Description

Fixes #1261

## Changes

This PR re-engineers how users interact with the policy parameter editor component. 

As a workaround due to the limitations inherent within the API, specifically how it currently stores custom policies, this code chooses an arbitrary year (2100) to represent "forever," storing this as a constant. It operates from the assumption that any policy defined to end on Dec. 31, 2100, is meant to occur "forever," altering the UX to reflect this assumption. 

With this PR, when a user loads the parameter editor, the default period displayed to a user over which a parameter change will be made is the current year "onward," representing 2100-12-31 behind-the-scenes. When a user clicks on the displayed year, which is now shown to the left of the value, an editor component with two tabs is shown. The left tab, labeled "yearly," edits a parameter from Jan. 1 of the first year to Dec. 31 of the final, mirroring the behavior of many of the users looking to select custom dates. The years displayed are restricted to the first year present within the respective country's metadata to 10 years from the last year in the metadata (i.e., currently Dec. 31, 2035 for US and Dec. 31, 2037, I believe, for UK).

For the remaining small group of users looking to select specific dates, the "advanced" tab provides an editor akin to what the app currently has, allowing custom date selection. Both the "yearly" and "advanced" editors use an Ant Design `Popover` instead of a collapsible component to prevent forcing the display chart downward, as well as due to the fact that the default collapsible component is a block element, while this altered date setter is intended to sit inline with the existing value setting component.

This PR also changes the right-hand sidebar to properly display "forever" dates, showing "onward" in these cases, as well. Finally, it refactors the `ParameterEditor` component by breaking out its individual `ValueSetter` and `PeriodSetter` components in order to allow for easier editing of the logic of either of these sub-components in the future. 

At the moment, I'm hoping to gain feedback on the visuals of this component. In particular, should the date setter look more like a button? I have tried a few different permutations visually, each of which came out worse than the current setup:
* Manually setting a border on hover leads to visual bugs due to the (faster) CSS-based hover pseudo-class Ant Design uses
* Adding a ":" after the date makes it feel slightly less obvious that it's a button
* Right-aligning the date leaves a strange-looking gap to the left
* Adding a gear or pencil icon beside the date looks unbalanced and strange

## Screenshots

https://github.com/PolicyEngine/policyengine-app/assets/14987227/cce39e78-2df3-45a1-b013-03ca2fba5f43

## Tests

This PR does not yet contain tests. These will be written following confirmation on the component's visuals.

### DIFF
--- a/src/data/constants.js
+++ b/src/data/constants.js
@@ -2,8 +2,8 @@
 // if a reform is applied into perpetuity; this should be replaced
 // with more substantive changes to how parameter changes work to 
 // allow for application into perpetuity, if that is desired behavior
-export const defaultForever = "2100"
+export const defaultForeverYear = "2100"
 
 export const defaultYear = new Date().getFullYear();
 export const defaultStartDate = defaultYear.toString().concat("-01-01");
-export const defaultEndDate = defaultForever.toString().concat("-12-31");
+export const defaultEndDate = defaultForeverYear.toString().concat("-12-31");
--- a/src/data/constants.js
+++ b/src/data/constants.js
@@ -1,8 +1,8 @@
 // The below is used to make parameter editor components appear as
 // if a reform is applied into perpetuity; this should be replaced
-// with more substantive changes to how parameter changes work to 
+// with more substantive changes to how parameter changes work to
 // allow for application into perpetuity, if that is desired behavior
-export const defaultForeverYear = "2100"
+export const defaultForeverYear = "2100";
 
 export const defaultYear = new Date().getFullYear();
 export const defaultStartDate = defaultYear.toString().concat("-01-01");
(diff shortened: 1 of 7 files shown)


nikhilwoodruff (approved):

anth-volk:
@MaxGhenis Would you mind looking over this and requesting any changes you'd like before I write tests and re-request review?

nikhilwoodruff (approved):

Issue #1555, opened 2024-04-03
Issues with author filtering when at bottom of Research page


anth-volk:
If a user is at the bottom of the Research page and decides to remove an author with many articles (e.g., Nikhil Woodruff) from the left-hand filter bar, they're jammed downward into the footer, instead of a more graceful handling of this situation. This is visible in the video below. Ideally, the user would be positioned at the last post that is now visible.

https://github.com/PolicyEngine/policyengine-app/assets/14987227/d5d10771-40f0-4b37-99ad-f496ed171e6f

anth-volk:
@NishantSatere separately logged this behavior in #1573, saying the following:

Current Behavior:
When a user is on the Research page and deselects an author from the filter options at the bottom of the page, the page automatically redirects the user to the bottom (footer) of the Research page. This behavior forces users to scroll back to the top to see the newly filtered content, which can be inconvenient and leads to a poor user experience.

Steps to Reproduce:

    Go to the Research page.
    Scroll down to the bottom of the page to access the author filter options.
    Unselect any author / location / topic from the filter checkbox options.
    The page refreshes and automatically scrolls down to the bottom (footer).

He also provides a video over there.

PR #1572, opened 2024-04-07
Issues with author filtering when at bottom of Research page #1555

NishantSatere:
## Description

I have a suggestion: instead of redirecting to the last post, it should redirect to new posts filtered by the user, which I have implemented."

## Screenshots

this is demo

https://github.com/PolicyEngine/policyengine-app/assets/100207648/df52cd94-c96d-4619-bf72-8a006e14227d

## Tests


MaxGhenis:
Please file an issue describing the current and proposed behavior in greater detail so we can discuss it. Thanks!

PR #1523, opened 2024-03-29
migration: migrated antd from v4 to v5

dotslashbit:
## Description

fixes #1479 

- Updated package.json and index.js file

Issue #1504, opened 2024-03-26
Move items from `redesign`'s `data` folder to `data` folder within `src` and rename the existing `data` folder in `src` to something more appropriate


anth-volk:
Ideally, everything in `src`'s current `data` folder should be in a new folder that better describes what it contains, since it doesn't contain data, it contains functions that manipulate data

PR #1576, opened 2024-04-08
Change Switch color to Blue 

### LINKED ISSUE #1531: Background color of Ant Design `Switch` components should be changed from Ant Design's default blue to our blue
For example, in this situation:

Screen Shot 2024-04-01 at 9 11 37 PM

The slider should use the dark blue that we use throughout. Not sure how this would work without looking through the documentation, but it should be relatively straightforward?

czhou578:
## Description
Fixes #1531

## Changes

Changes the color of the ant design switch components to (#2C6496) which is specified as "BLUE" in the colors.jsx file.

## Screenshots

image

## Tests

No extra tests have been added


czhou578:
@anth-volk , could you take a look at the BiPanel.jsx to see how it is being used? I have already applied the blue color to all the other files with switch elements. Thanks!

anth-volk (changes requested):
Thanks for this @czhou578, this is very helpful. Unfortunately, there's a minor bug in this, as visible in the video below. If a value is already selected and the `Switch` component is already true for the user, it still displays the default color. I'm guessing this is because your changes manually manipulate the design via JS. Did you happen to find any way to set this as a property within the Ant Design documentation? That said, there may be a way to set this imperatively, as well.

`BiPanel` is a remnant of a previous version of the site and should be deleted as part of #1504. That said, please feel free to keep the code you added.

czhou578:
Thanks for clarifying what BiPanel.jsx was doing. I guess I'll just leave that file in there. In the Ant documentation, I could not find a specific property that exists which allows me to set the color of the switch element. By the way, I don't see the video that you mentioned in your comment...

anth-volk:
Haha that would be because I didn't add it. Here it is:

https://github.com/PolicyEngine/policyengine-app/assets/14987227/5777541d-1a82-44c9-a1b6-ba981aca22a6

czhou578:
@anth-volk, I don't believe I can imperatively set the switch color for the PolicyRightSidebar. If a user checks the switch and then refreshes the page, then the state tracking the checked state (default value is false) would stay false, even though the switch is already checked. I'm thinking we can use the localStorage on the browser so that we can persist the checked state even through refreshes, but I'm not sure if that's acceptable. What do you think?

anth-volk:
I'm thinking the best thing we could do is upgrade to Ant Design v5, then implement their context provider, as looking through their docs, I believe that actually sets style changes via CSS and does so in a more comprehensive way.

At the moment, the upgrade is being done in #1523 by a different contributor, though I think it's been about a week since they've worked on it. I think we could see if they continue, but if not, would you be willing to continue that down the road? Then perhaps this (#1576) could be resolved via the context provider.

czhou578:
Makes sense. I will be willing to work on that issue. Maybe I'll give the other contributor another 1-2 days to make additional changes? If nothing happens by then, I can start working on that.

anth-volk:
Actually, why don't I reach out to him directly tomorrow morning?

anth-volk:
Just got confirmation from @dotslashbit that he'd be happy for you to take on the upgrade, @czhou578. This is currently open as PR #1523. Could you push the requested changes to his fork branch so that we can utilize the same PR?

czhou578:
@anth-volk I have tried to push my changes to his fork branch, but it gives me this error:

image

anth-volk:
Ah, I think you'd have to clone his fork, then push against the branch. 

Instead, then, could you try just pulling his changes in, then dropping yours on top and opening a new PR? Then we can close out his.

anth-volk:
Closing in light of #1584. In order to properly do this, we'll need to do the things mentioned in that PR first and migrate to Ant Design v5 first.

Issue #1442, opened 2024-03-04
The axis tick and hover numbers are uninformative for some household charts


abhcs:
In [this example](https://policyengine.org/us/household?focus=householdOutput.earnings&household=41940&region=us&timePeriod=2024&baseline=2&reform=48789), all values are `60` but the plot is not a flat line:

<img width="760" alt="Screenshot 2024-03-04 at 9 29 27 AM" src="https://github.com/PolicyEngine/policyengine-app/assets/31144719/69af219f-22da-44cd-be1e-6b057099e0f4">

PR #1583, opened 2024-04-10
Authenticated user experience

### LINKED ISSUE #588: Simple sign-in
Just so we have some initial version and have at least the capability to save user details, we should enable users to sign in to the site. This initial version doesn't need to have any actual functionality other than a sign-in ability.

### LINKED ISSUE #1509: Move modals to dedicated folder

### LINKED ISSUE #1557: Create page for authenticated users
This page should be visible if a user is logged in; otherwise, it should be inaccessible. It should contain a section showing users all policies that they've previously viewed, which will be stored in a table that should be accessible by GET request. There should also probably be some sort of minimal profile view on this page.

### LINKED ISSUE #1558: Add click event handling to add a policy that a logged-in user has viewed
This will require commensurate changes to the API before implementation

### LINKED ISSUE #1561: Add login/logout button
Perhaps this should be in the header bar? This will require some design confirmation

### LINKED ISSUE #1562: Add login/signup prompt on policy calculator run
This should be based off of #1537, with some obvious changes. Being that this is integrated with auth0, there's no need for the standard "log in/sign up" page, just the prompt. This component should also make some consideration for users without cookies enabled, though it's unclear what just yet.

### LINKED ISSUE #1564: Make navbar more consistent across breakpoints and with similar sites
Currently, I think there are three major drawbacks with the navbar:

1. Its height is 90px (very large) on desktop/tablet, but only 50px on mobile
2. While it's mainly written against flexbox, liberal use of margins makes maintenance difficult
3. It can really only fit a few options as written

Its height is particularly significant, since most sites place theirs between 60px and 80px, and its large size is what necessitated setting a custom value on mobile. I think this should all be changed as follows:

1. Navbar height for all breakpoints should be set to 64px; Pieter Heyman has a good explanation of why [here](https://blog.prototypr.io/a-quest-for-the-ideal-navbar-height-4d3898a41f2), but this is also what Material Design uses for its header bars
2. The component should better organize items by using a nested flexbox approach, as opposed to manual margin settings
3. Text height should be dropped to 16px

I will likely need to do these as part of #1561, as merely adding it to the existing bar produces a part that is extremely cluttered and forces poorly-sized wrapping on the large feature button that we utilize

anth-volk:
## Description

Fixes #588.
Fixes #1509.
Fixes #1557.
Fixes #1558.
Fixes #1561.
Fixes #1562.
Fixes #1564.

## Changes

This PR creates an authenticated user experience. It integrates auth0 to allow users to connect with the application via their choice of one of three auth providers. After login, users have access to a user profile page, which stores all policy calculator simulation runs that they have run while signed in, as well as any run within the last two weeks if not signed in, so that users who use the app a few times and decide they want to create an account are able to still access their previous runs.

This PR also refactors the navbar to fit with more common standards and ensures that its size is consistent between mobile and desktop. The PR moves modals to a new `modals` folder, as well.

## Screenshots

The video below provides a brief overview of the new navbar, the login experience, the profile page, and how a user without an account can save their policies by logging in.

https://github.com/PolicyEngine/policyengine-app/assets/14987227/f26b6c5c-a55c-40d2-b743-f34f33477af8

## Tests
While the API structures that this PR are reliant upon have full test coverage, the front-end portions of this do not, for three reasons:
* Most of the front end additions in this PR merely consume the back end, so while front-end tests in these places are important, they're not as vital as the accompanying back-end ones
* Anything to do with connecting with auth0 is either difficult or not possible to test, and any future tests should merely make the assumption that auth0's API works, as it is an external, consumed API that we do not control
* These changes are intended to be tested by the team post-push, but pre-official launch


src/posts/postTransformers.js
```diff
@@ -0,0 +1,93 @@
...
+const postsSorted = posts.sort((a, b) => (a.date < b.date ? 1 : -1));
+
+for (let post of postsSorted) {
+  post.slug = post.filename.substring(0, post.filename.indexOf("."));
+}
+
+const tags = postsSorted.map((post) => post.tags);
+const uniqueTags = [...new Set([].concat(...tags))].sort();
+
+const locationTags = uniqueTags.filter((tag) =>
```
PavelMakarchuk, line 12: should we also use Canada as we are developing the model?

anth-volk: In the location labels, you mean? This file is actually just a reimplementation of the old `Posts.jsx`, meaning it's live in the app since the code was merged today, so I'll open a new issue to add that. Any need for Nigeria, Israel, or others?

nikhilwoodruff:
Nearly there- thanks! Could we:

* Use this icon instead of the coffee mug. 
image

* Expand the width of the cards so they're (at highest) 1:1 height-width ratio.
image

nikhilwoodruff:
And finally could the links point to the overview economic impact page (just using the focus= parameter).

anth-volk:
@nikhilwoodruff Could you describe a bit more in detail what you're thinking regarding card height and width? And how should your recommendation work across breakpoints?

Otherwise, thanks for the feedback, will work on. There remains one more challenging bug that I'll need to discuss - the fact that, by design, user accounts are meant to access all policies on any country version of the site, but we don't have access to other countries' metadata objects at that time (e.g., when a user with exclusively US simulations visits the UK site). 

I also have a bit more work to do on displaying modals properly on simulation runs, and I'll have to make sure that these changes work correctly in light of recent API changes (#1442, etc.).

nikhilwoodruff:
@anth-volk-

On the height ratio, I just meant that we should cap the height ratio so it is no more than 100% of the width of the card: cards shouldn't be taller than they are wide.

On the more challenging bug- don't worry about it and let's just hide the non-country cards. Users can easily just go to policyengine.org/[country], this feature is in the early stages and this is a use-case that applies to very few people, mostly inside PolicyEngine.

nikhilwoodruff (approved):

Issue #912, opened 2023-11-28
Calculator child input and marriage input buttons are rounded


anth-volk:
<img width="697" alt="Screen Shot 2023-11-28 at 2 17 27 PM" src="https://github.com/PolicyEngine/policyengine-app/assets/14987227/8e36f25b-e24d-4be1-9eda-c6258a6a2991">

They should ideally be squared like all other buttons on the site

MaxGhenis:
This is true of all categoricals right?

I think we should go with a standard like the antd radio buttons instead, to make it more obvious that it's a choice: https://ant.design/components/radio

Specifically this is closest:
<img width="237" alt="image" src="https://github.com/PolicyEngine/policyengine-app/assets/6076111/dee85535-2f1e-416f-83c0-d19fd30533bb">

```
optionType="button"
buttonStyle="solid"
```

anth-volk:
Good point, could improve accessibility as well. 

And I suppose it is true for all categoricals, but since link buttons are squared, I wonder if maybe we should use an antd standard, but squared? Or minimally rounded, as in the image you sent, at the least.

MaxGhenis:
This is a bigger conversation but I'm partial to going with a design standard like Ant for all buttons. I think this would improve usability, accessibility, and maintainability. For example the button color swiping across makes it feel sluggish, especially when moving a cursor across multiple buttons. I think there are good reasons for the kinds of choices designers at Ant and elsewhere adopt when creating standards.

maurobusso:
I personally think in this situation with a big squared button below, two rounded one on the top would make the page look stylistically a bit messy. But this is just my opinion. :)

daemon-reconfig:
can I take this issue?

anth-volk:
@daemon-reconfig Sure. Just a few things to note:

* As described in the README, please open a PR off of a fork, not the cloned repo
* When opening the PR, please add "Fixes #issue number" in the description
* This should be redone as an Ant Design radio component, as described by Max in the description above. That said, we'd ideally like the background color of a selected item to be "BLUE_PRIMARY", available as a constant in `src/redesign/style/colors`. If this can't be implemented, it may be better to wait until we upgrade to Ant Design v5.

daemon-reconfig:
sure thanks for letting me know. I am terribly sorry for this late reply

daemon-reconfig:
[...]

PR #1600, opened 2024-04-12
Changed to ant design radio buttons

daemon-reconfig:
## Description

this fixes: [#912 ](https://github.com/PolicyEngine/policyengine-app/issues/912)

## Changes

Changed to ant design radio buttons 

## Screenshots

Screenshot 2024-04-13 at 2 15 37 AM


anth-volk:
I'm noticing that the border radius on this isn't quite squared off. Could you set it to 0? When we migrate to Ant Design v5, this will become much rounder, too much for our design.

anth-volk:
Could you run `make format` and push the results? We do require linting before merging in order to make for better code quality.

daemon-reconfig:
sure i will do so

daemon-reconfig:
Screenshot 2024-04-13 at 10 34 29 AM

here fixed

daemon-reconfig:
Screenshot 2024-04-13 at 10 59 36 AM

it gives this while linting

MaxGhenis:
Try `pip install black` first (see https://github.com/PolicyEngine/policyengine-app/issues/1602)

daemon-reconfig:
> Try `pip install black` first (see #1602)

thanks!

daemon-reconfig:
@anth-volk could you review this PR and merge it if it suits ?

anth-volk:
I will try to do so today, but am not sure if I'll be able to. It didn't appear that you'd run the linter, which is why I held off, but the code does pass all tests, so perhaps you had?

daemon-reconfig:
yes had linted the code

package.json
```diff
@@ -14,8 +14,9 @@
     "@uiw/codemirror-theme-github": "^4.21.25",
     "@uiw/react-codemirror": "^4.21.25",
     "@uiw/react-textarea-code-editor": "^3.0.2",
-    "antd": ">4, <5",
+    "antd": "^4.24.16",
     "axios": "^1.6.8",
+    "black": "^0.3.0",
```
anth-volk, line 19:
```suggestion
```

anth-volk: This package is different from the Python `black` linting package and should be uninstalled via `npm uninstall black` before recommitting

anth-volk (changes requested):
Thanks for this, @daemon-reconfig. Just a couple of things:

1. If you run the code, you'll notice that the borders between the buttons are poorly defined, since they're a shade of gray. This is probably because of how Ant Design creates these borders. Could you instead iterate over the labels and generate individual `Radio.Button` elements to sit within one `Radio.Group`, and have any non-selected buttons have their own `style={{}}` prop that contains `border: "1px solid style.colors.WHITE"`?
2. This component currently doesn't manage styling statefully, instead always just making the background of the first element in the list blue and keeping the others gray. Could you use a standard state handling structure within `Radio.Group`'s `onChange` to save something (perhaps an index) to a stateful variable, then conditionally apply a background color to each `Radio.Button` within the `Radio.Group`? This will probably require a function that looks something like the following:

```
function handleChange(may need args here, depending on how Ant Design passes events) {

  Check if PolicyEngine's RadioButton's onChange prop is a function

  If so, run that function

  No matter what, update state to make this a controlled component
}
```

I've yet to actually test to make sure values are passed properly, but they appear to be.

Thanks again.

daemon-reconfig:
sure I will do the following changes.

daemon-reconfig:
if I uninstall that black I get error while linting

daemon-reconfig:
@anth-volk could you please elaborate what you want me to add in it

daemon-reconfig:
should I map radio.buttons? and 0th index will be the selected one the border of non selected should be changed to white
is this what you are talking about?
or I can change the border color of the non selected button to white

anth-volk:
The issue with `black` is partly on us, and it's being fixed in #1623. The `black` that you installed is a JavaScript package, but what contributors actually need is the Python package `black`, so if you could run `make install`, then `pip install -U black` for the time being, that should fix your issue. If that fails, please let me know.

anth-volk:
And yes, the borders should be white for all non-selected buttons, but what you have now for the selected button works. You should probably place `Radio.Group` into the main return statement, since you only need one of those, then map `Radio.Button`, but adding event change handling, akin to building a standard fully controlled React component. I could go more into detail if you're unfamiliar with this.

daemon-reconfig:
I had done pip install black when I installed it @MaxGhenis told me to.
an sure I will add an event handler for the buttons but I didn't get the thing about radio.button

daemon-reconfig:
1. I am so sorry about it I did run `npm i black`
I will uninstall it.

2. I do know why we need states but I didn’t get why we should add `Radio.Button` as in the given reference there was no mention of it
https://ant.design/components/radio
But since you asked me to I will add them to it
what I did get is
• Adding a `Radio.Button` so that the first button would be selected which will have normal border- blue given by ant design itself 
• the second button which would be unselected should have white borders
• above two points should be achieved using states which will change the colors of the first button if it is unselected, i.e, second button was selected

my opinion: since we would not just be targeting two buttons using `Radio.Button` might not do it although we can map an array which will give us `Radio.Button`.

I deeply apologise if I caused any misunderstanding. 
I am thankful for your helpful messages and you have helped me learn a lot of things.

anth-volk:
1. Not a problem at all
2. Ant Design's documentation, unfortunately, isn't always amazing. I'm thinking `Radio.Button` would be better because with just the group, it seems a bit more difficult to manage state and to conditionally format the buttons. If you can get it to work with just `Radio.Group`, please feel free, that was just my read of the documentation.

anth-volk:
Also, in the future, if you're still working on an issue, please feel free to convert it to "draft" by clicking "Still in progress? Convert to draft" in the top right. Typically, if you close a PR, it looks like you no longer want to work on the issue.

daemon-reconfig:
sure I will add it to draft. 
also I think we can make do with `Radio.Group` and I will start working on implementing `Radio.Button` in it. as it may take time since I will have to research a bit into it.
or we can either go with more time taken but the end result would be getting a `Radio.Button`

anth-volk:
We're open to however you'd like to implement it, whichever way you think will work best.

daemon-reconfig:
sure will open a PR for it on Sunday. I have an exam tomorrow

PR #1622, opened 2024-04-17
correct rounding of percentages in parameter history hovercard

masterismail:
## Description

noticed that the `Math.round` was causing the rounding off to next digit 

## Changes

used `Math.floor` instead

## Screenshots

Screenshot from 2024-04-18 01-19-40

LGTM

## Tests

have run `make test` and `make format` , both of them passing.

### DIFF
--- a/src/api/variables.js
+++ b/src/api/variables.js
@@ -245,7 +245,7 @@ export function getPlotlyAxisFormat(unit, values, precisionOverride) {
     const r = range();
     const s = unit === "/1" ? 100 : 1;
     const d = (r[1] - r[0]) * s;
-    return d > 10 ? 0 : d === 0 ? 1 : 1 - Math.round(Math.log10(d));
+    return d > 10 ? 0 : d === 0 ? 1 : 1 - Math.floor(Math.log10(d));
   };
 
   const isCurrency = Object.keys(currencyMap).includes(unit);


anth-volk (changes requested):
Thanks for your contribution, @masterismail, and apologies for the delay in reviewing. Unfortunately, this doesn't actually solve the issue Max highlighted. As you can see in the image below, when I run the code, the small blue box that pops up on hover still displays "5%" instead of "4.67%", which is what we're looking for.

Screen Shot 2024-04-19 at 6 15 13 PM

anth-volk (approved):
Thanks @masterismail, this looks good!

Also, just as advice, in the future, you may consider following a slightly different flow when opening PRs:
* Fork the repo that you want to contribute to
* Keep branch master or main in sync with the original repo by pulling the original repo's code at times; typically (if the original repo is called "upstream" by Git) this means running git pull upstream master, then git push origin master to sync the code to your local repo
* Create branches on your fork off of your master or main branch. Periodically, if you're working on something for a while, you might also run git rebase master within your feature branch to sync your code with any new changes.
* When you're done with the feature or bug fix, open the PR from your branch to the live main or master.

If you prefer using a more visual editor, you can add the VS Code GitHub extension and use the [Start Working on This Issue](https://code.visualstudio.com/blogs/2020/05/06/github-issues-integration#_working-on-issues) feature, which will automate that process.

This just ensures that we don't run into merge conflicts down the road, especially if you contribute for a longer period.

Thanks so much!

PR #1623, opened 2024-04-17
Added pip install black to makefile

### LINKED ISSUE #1602: Add `pip install black` to `make install`

czhou578:
## Description

Fixes #1602 

## Changes

I added the command "pip install black" to the makefile so that running make install will automatically install the dependency.

## Tests
No new tests were added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 install:
 	npm ci
-	pip install black
+	pip install -U black
 build:
 	npm run build
 debug-no-lint:
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 install:
 	npm ci
+	pip install black
 build:
 	npm run build
 debug-no-lint:


anth-volk (approved):
Thanks so much for this, @czhou578, this is great to have. At the moment, somewhere in the GitHub Actions (either push or PR, I'm not sure which), we actually install `black` at a particular version in order to check linting. Could you modify this to use whatever version it is that is specified there, test it to make sure it works, and push those changes? We have had instances where someone's installed version doesn't match that version, leading to some confusion. Otherwise, thank you very much for this!

czhou578:
I read through the Black documentation and it seems like the pip command should be installing the stable version, since they give an alternative way to install the latest changes from github if the user wants to. I tried doing ```pip install black@stable``` and the command gave an error. 
image

anth-volk:
That makes sense, but I believe if you already have a version set up on your machine, it will just use whatever version you have - this probably won't be an issue for front-end devs who contribute, but for anyone who comes from the Python development space, it could cause issues. Could you instead change the script to 

```
pip install -U black
```

This will install whatever the latest stable version pip can find is.

anth-volk (approved):
Thanks so much for this, @czhou578, this works like a charm.

PR #1628, opened 2024-04-19
More safely handle conditional require statements in blog post

### LINKED ISSUE #1589: More safely handle conditional `require` statements in blog post
Currently, in the `BlogPost.jsx` component, we use a conditional `require` to load blog images. However, the conditional only checks if a blog image is specified, not if it's present. I'm not sure if this is technically feasible, but it'd be great to ensure that a failing `require` statement returns some sort of empty "broken file" block, akin to what standard JS-free HTML image loading would do, as opposed to killing the entire app.

czhou578:
## Description

Fixes #1589 

## Changes

Added a conditional try catch block to check for the existence of a blog post image. This is done in order to handle the nonexistence of a picture in a more graceful manner. If an image does not exist, then a ```FileImageOutlined ``` icon from Ant Design is displayed at the top.

## Screenshots
image

## Tests

No new tests added.

### DIFF
--- a/src/pages/BlogPage.jsx
+++ b/src/pages/BlogPage.jsx
@@ -57,7 +57,6 @@ export default function BlogPage() {
   ) : (
     <FileImageOutlined />
   );
-  console.log(imageUrl);
   const file = require(`../posts/articles/${post.filename}`);
 
   const [content, setContent] = useState("");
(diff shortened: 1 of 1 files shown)


czhou578:
@anth-volk , how big do you want the missing pictures icon to be?

czhou578:
@anth-volk, how about this?

image

anth-volk:
What about this - in place of the image, pop in a div with a height of 300px, the same width as the image (is it 100%?), and a 1px gray border. Flexbox it, and place the icon (maybe, like, font-size 32px?) into the bottom-right corner. It's similar to what you had been thinking with the broken image icon, I think.

czhou578:
Something like this?

image

anth-volk:
That looks perfect to me

czhou578:
Ok. I've pushed the changes.

anth-volk:
Perfect, will review Monday. Thanks so much for this, @czhou578.

anth-volk (approved):
This is perfect, @czhou578, thank you very much!

Issue #1629, opened 2024-04-19
More safely handle conditional require statements for blog article images on homepage, main Research page


anth-volk:
Related to #1589, it would be great to do what @czhou578 is doing in #1628 for both the homepage and the main Research page.

anth-volk:
@czhou578 Would you also be willing to take this issue? This is targeted at the Research page and homepage.

czhou578:
Sure. I can take this issue.

czhou578:
@anth-volk, do you want me to do checks on these images as well?
![image](https://github.com/PolicyEngine/policyengine-app/assets/42430035/53f5d28c-f991-4b31-98d0-edffd8104e17)
![image](https://github.com/PolicyEngine/policyengine-app/assets/42430035/28daf282-5b49-46ad-8eab-a445b728ca40)

anth-volk:
Sure, please do so.

PR #1636, opened 2024-04-23
Added conditional require statements for blog images on Home and Research pages.

czhou578:
Fixes: #1629 

## Changes

Added conditional require statements for blog images on Home and Research pages. I also added a conditional check for some additional images on the Home page, as demonstrated through the following screenshots.

## Home
image
image
image

## Research

image

## Tests

No new tests were added.

### DIFF
--- a/src/redesign/components/HomeBlogPreview.jsx
+++ b/src/redesign/components/HomeBlogPreview.jsx
@@ -409,11 +409,22 @@ export function FeaturedBlogPreview({ blogs, width, imageHeight }) {
 export function MediumBlogPreview({ blog, minHeight }) {
   const displayCategory = useDisplayCategory();
   const countryId = useCountryId();
-  const imageUrl = blog.image
-    ? require("../../images/posts/" + blog.image)
-    : require("../../images/placeholder.png");
   const slug = blog.filename.split(".")[0];
   const link = `/${countryId}/research/${slug}`;
+
+  const handleImageLoad = (path) => {
+    // Try to load the image
+    try {
+      return require("../images/p osts/" + path);
+    } catch (error) {
+      // If the require fails, return the fallback image
+      console.error(`Failed to load image at ${path}:`, error);
+      return "";
+    }
+  };
+
+  const imageUrl = blog.image ? handleImageLoad(blog.image) : "";
+
   return (
     <div
       style={{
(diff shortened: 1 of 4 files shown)


anth-volk:
Here again, could you instead treat the icon like a thumbnail and place it in the bottom-right corner of a div that fills the space where the image would be?

czhou578:
Is this acceptable?

image
image

src/redesign/components/HomePolicyEngineFeatures.jsx
```diff
@@ -6,7 +6,30 @@ import decileChartScreenshot from "../images/home/decile_chart_screenshot.png";
 import ShowcaseItem from "./ShowcaseItem";
 import useCountryId from "./useCountryId";
 
+function ImageComponent(imagePath) {
+  let imageSrc;
+  try {
+    // Attempt to import the image dynamically
+    imageSrc = require(`../images/home/${imagePath}`).default;
+    console.log(imageSrc);
```
anth-volk, line 14: Could you remove this console log?

src/redesign/components/HomeBlogPreview.jsx
```diff
@@ -323,14 +324,25 @@ function BlogTags({ tags }) {
   );
 }
 
+const handleImageLoad = (path) => {
+  // Try to load the image
+  try {
+    return require("../images/posts/" + path);
```
anth-volk, line 330:
```suggestion
    return require("../../images/posts/" + path);
```

anth-volk: We're gradually trying to deprecate the `/redesign` folder, including its images, and the whole thing will be fully removed in an upcoming fix, but for now, this should reference the standard images folder, not the one inside `/redesign`

src/redesign/components/ShowcaseItem.jsx
```diff
@@ -44,16 +46,38 @@ export default function ShowcaseItem({
...
+          />
+        </div>
+      ) : (
+        <img
+          src={image}
+          width={displayCategory === "desktop" ? 400 : "100%"}
+          height="100%"
+          style={{
+            objectFit: "contain",
+            // boxShadow: "0px 0px 10px rgba(0, 0, 0, 0.9)",
```
anth-volk, line 76: Could you remove the commented styling?

anth-volk (changes requested):
Thanks for this, @czhou578! This all looks great, just had some minor changes, and then I think this will be merge-able. Thanks again for your help with this.

czhou578:
Fixed!

anth-volk (approved):
Thanks for the changes @czhou578! This is perfect and will help increase the stability of the application when errors occur.

Issue #1640, opened 2024-04-25
EPMT logo on testimonials doesn't properly display


anth-volk:
If you visit https://policyengine.org/us/testimonials, the logo for End Poverty, Make Trillions doesn't properly display.

PR #1641, opened 2024-04-26
Use Enhanced CPS when specified for Reproduce in Python Microsimulation call

### LINKED ISSUE #1624: Use Enhanced CPS when specified for Reproduce in Python Microsimulation call
Currently the Reproduce in Python for microsimulations uses this code for all simulations:
```python
baseline = Microsimulation(reform=baseline_reform)
reformed = Microsimulation(reform=reform)
```
When the Enhanced CPS is toggled, it should instead use:
```python
baseline = Microsimulation(reform=baseline_reform, dataset="enhanced_cps_2022")
reformed = Microsimulation(reform=reform, dataset="enhanced_cps_2022")
```

shaunakwagh:
## Description
Fixed #1624

## Changes

 Reproduce in Python gives a different code when enhanced CPS is enabled

 
## Screenshots
Before
Screenshot 2024-04-26 at 10 18 57 AM
After
Screenshot 2024-04-26 at 10 19 12 AM

## Tests
No test added

### DIFF
--- a/src/data/reformDefinitionCode.js
+++ b/src/data/reformDefinitionCode.js
@@ -206,7 +206,7 @@ export function getImplementationCode(type, region, timePeriod) {
   }
 
   const isCountryUS = US_REGIONS.includes(region);
-  //"reform=baseline_reform, dataset=\"enhanced_cps_" + timePeriod+"\""
+  
 
   return [
     "",
--- a/src/data/reformDefinitionCode.js
+++ b/src/data/reformDefinitionCode.js
@@ -206,7 +206,6 @@ export function getImplementationCode(type, region, timePeriod) {
   }
 
   const isCountryUS = US_REGIONS.includes(region);
-  
 
   return [
     "",
--- a/src/data/reformDefinitionCode.js
+++ b/src/data/reformDefinitionCode.js
@@ -213,13 +213,13 @@ export function getImplementationCode(type, region, timePeriod) {
     `baseline = Microsimulation(${
       isCountryUS
         ? region === "enhanced_us"
-          ? `reform=baseline_reform, dataset="enhanced_cps_${timePeriod}"`
+          ? `reform=baseline_reform, dataset="enhanced_cps_2022"`
           : `reform=baseline_reform`
         : ""
     })`,
     `reformed = Microsimulation(${
       region === "enhanced_us"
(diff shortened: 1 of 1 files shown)


anth-volk:
I believe you have two sets of changes in this branch, one which is for #1624, and one which is for #1640. Could you separate these and ensure that this PR is just for #1624?

shaunakwagh:
i removed the changes for #1640

anth-volk:
Will review a bit later today

shaunakwagh:
any updates ?

anth-volk:
My apologies, somehow I missed this. Will review in a few.

src/data/reformDefinitionCode.js
```diff
@@ -211,13 +211,21 @@ export function getImplementationCode(type, region, timePeriod) {
...
+        : ""
     })`,
-    "reformed = Microsimulation(reform=reform)",
-    `baseline_person = baseline.calc("household_net_income",`,
-    `    period=${timePeriod || defaultYear}, map_to="person")`,
-    `reformed_person = reformed.calc("household_net_income",`,
-    `    period=${timePeriod || defaultYear}, map_to="person")`,
+    `reformed = Microsimulation(${
+      region === "enhanced_us"
+        ? `reform=reform, dataset="enhanced_cps_${timePeriod}"`
```
anth-volk, line 222:
```suggestion
        ? `reform=reform, dataset="enhanced_cps_2022"`
```

While your assumption that this should change by year makes a lot of sense, I believe it is actually meant to map to 2022 specifically. If you try a different year in your version of the code and run the output in Google Colab, for example, you receive the following error:

```
ValueError: Dataset enhanced_cps_2023 not found. Available datasets: ['cps_2019', 'cps_2020', 'cps_2021', 'cps_2022', 'enhanced_cps_2022', 'calibrated_puf_extended_cps_2022', 'poverty_tracker', 'puf_2022', 'puf_2015']
```

anth-volk (changes requested):
Thanks for this, @shaunakwagh. While I appreciate that you anticipated the code incorrectly setting the year, the enhanced CPS flag actually has to be set to 2022. Could you change that then re-request review? After that, this should be good to go.

anth-volk (approved):
Thanks so much, @shaunakwagh! This is now what we're looking for.

Issue #1642, opened 2024-04-26
Remove deprecated Twitter code


anth-volk:
`index.html` contains a deprecated `<script>` tag that connects with Twitter in order to load a Twitter component that was later moved. This should be removed, as it probably has an impact on load time.

anth-volk:
Fixed by #1645.

PR #1645, opened 2024-04-27
 Remove deprecated Twitter code #1642

LW1117:
## Description

 Remove deprecated Twitter code #1642

## Changes

 Removed script tag with deprecated Twitter code in index.html

## Screenshots
Screenshot 2024-04-27 101429

## Tests

The change has been tested on running at local host.


anth-volk (approved):
Thanks so much, @LW1117! This definitely cuts down on the number of fetches during load. Approved.

anth-volk:
Thanks again for this, @LW1117. In the future, when adding an issue number, could you write it like so:
```
Fixes #1642 
```
In your current format, GitHub won't recognize it as being fixed, only referenced.

PR #1668, opened 2024-04-30
Added conditional redirect_uri if running in debug mode

### LINKED ISSUE #1663: Conditionally define redirect URI for Auth0ProviderWithNavigate component
At present, the component always redirects to the live web address, but when a contributor is running locally, the redirect URI should instead be set to localhost to allow for proper testing. The best way to do this might be some sort of edit to the makefile, perhaps by setting some sort of environment variable within the `make` recipe that indicates that the app is in debug mode, and then conditionally seeking that value within `Auth0ProviderWithNavigate`.

anmolchhabra21:
## Description

Fixes #1663 
## Changes

Changed redirect_url as per environment to point to local for developers(debug command), and policyengine for other cases.

## Screenshots
N/A

## Tests

Formatted and Tested 18/18 passed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,21 @@
+REACT_APP_DEBUG ?= false
+
 install:
 	npm ci
 	pip install -U black
+
 build:
 	npm run build
+
 debug-no-lint:
 	ESLINT_NO_DEV_ERRORS=true npm start
+
 debug:
-	npm start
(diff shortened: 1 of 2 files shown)


anth-volk (approved):
This is perfect, @anmolchhabra21, thanks for your contribution!

PR #1680, opened 2024-05-01
Added escape key handler

### LINKED ISSUE #1656: Add `Enter` and `Esc` button event listeners to username change component on UserProfilePage
At the moment, you must click either the checkbox to save a new username or the X to exit, you can't use Enter or Esc. This should change.

czhou578:
## Description

Fixes #1656 

## Changes

I added an event handler to handle the case when the user presses the escape key when editing their username on the UserProfile page. Before I started editing, it seemed like the "Enter" button was already working, strangely, so I went straight to handling the escape.

## Screenshots

[

https://github.com/PolicyEngine/policyengine-app/assets/42430035/c2c08405-c319-42eb-bb72-ef83d2d46c87

](url)

## Tests

No new tests were added.

### DIFF
--- a/src/pages/UserProfilePage.jsx
+++ b/src/pages/UserProfilePage.jsx
@@ -608,6 +608,11 @@ function UsernameDisplayAndEditor(props) {
             defaultValue={accessedUserProfile.username}
             size="small"
             onPressEnter={handleSubmit}
+            onKeyDown={(e) => {
+              if (e.key == "Escape") {
+                setIsEditing(false);
+              }
+            }}
             onChange={handleUpdate}
             style={{
               height: "1.3rem",
--- a/src/pages/UserProfilePage.jsx
+++ b/src/pages/UserProfilePage.jsx
@@ -339,9 +339,9 @@ function UserProfileSection(props) {
       }}
     >
       {dispState === STATES.OWN_PROFILE &&
-        isAuthenticated &&
-        user &&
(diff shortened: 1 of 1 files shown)


czhou578:
@anth-volk, seems like some things in the jest files are now deprecated. How can this be fixed?

anth-volk:
@czhou578 It appears the full failure is actually due to an API timeout - unfortunately, I've yet to find any way of just fetching metadata once and passing it to all test files, and sometimes this timeout does occur in the components themselves. That said, there are a ton of errors for soon-to-be deprecated Jest features, which I'll open in a separate issue.

czhou578:
Ok. It seems like the CI tests passed already. I can take the issue of updating Jest, since I would like to improve my knowledge of unit testing. Is that ok?

anth-volk (approved):
Thanks for this, @czhou578! This does work as anticipated; technically, because of the way React works, if you're not actively in the input box, but the box is open, and you hit "Escape," nothing will happen, as React is only looking for the event on that particular item. A more comprehensive fix would've been to place a manual event listener (with a conditional check to make sure the box is open) in an effect hook, as that extends the event across the entire document. That said, I don't think this fix requires that level of quibbling, so I think this fix works well. Thanks again!

anth-volk:
And sure, please feel free to handle the updating of Jest

Issue #1699, opened 2024-05-02
SignupModal's header font is incorrect


anth-volk:
This is likely due to the fact that `App.css` only defines the font for h1 through h4 headers, and that modal uses h6. We should either change it to something like h4 or alter `App.css` to define the same font for all headers.

rhwd:
I made a PR for this issue, I would be pleased if you could review it. 
https://github.com/PolicyEngine/policyengine-app/pull/1700

PR #1700, opened 2024-05-03
Add common font between all headers

rhwd:
## Description

A small fix for  #1699

## Changes

Added the h5, and h6 elements to the existing CSS rule for headers

### DIFF
--- a/src/modals/SignupModal.jsx
+++ b/src/modals/SignupModal.jsx
@@ -37,7 +37,7 @@ export default function SignupModal(props) {
       <h6
         style={{
           paddingBottom: "16px",
-          fontWeight: "bold",
+          fontWeight: 500,
           fontSize: 20,
         }}
       >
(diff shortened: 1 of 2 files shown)


anth-volk (changes requested):
Thanks for this, @rhwd. This does do exactly what the issue says, but would you mind also changing the "fontWeight" property in the `src/modals/SigupModal` file from "bold" to "500" to match other fonts? The modal still looked incorrect and it appears that's also part of the problem.

rhwd:
Thanks for the heads up and instructions, @anth-volk. Just pushed a fix for this.

anth-volk (approved):
This is perfect! Thanks @rhwd. Will merge once the GitHub Actions finish.

In the future, we typically recommend that contributors fork our repo, then keep their `master` branch up to date with the live branch, making edits on a local feature branch. Instructions on how to do so are available in our README.

Issue #1690, opened 2024-05-02
Move the "Run the code below" line from the Reproduce in Python pages so that it sits above the code editor


anth-volk:
<img width="1440" alt="Screen Shot 2024-05-02 at 6 44 43 PM" src="https://github.com/PolicyEngine/policyengine-app/assets/14987227/12e1f169-1ffb-4225-98ff-5e94881167bf">

psankhe28:
I've made the required changes and created the PR https://github.com/PolicyEngine/policyengine-app/pull/1706
Please have a look at it

anth-volk:
Thanks @psankhe28, I'll be sure to check them out today.

psankhe28:
Please close this issue @anth-volk as the required changes are done.
Thanks!

anth-volk:
Thanks, fixed by #1690.

PR #1706, opened 2024-05-03
Changes in the position of Line

psankhe28:
## Description
#1690 

## Changes

Moved the "Run the code below" line from the Reproduce in Python pages so that it sits above the code editor 

## Screenshots

image

## Tests

17/17 passed


anth-volk:
Thanks for this @psankhe28. While it does look good, could you also just lower the bottom padding on the line so that it matches the amount below "Include earning variation"? Otherwise, I think it'll be good to go after that.

psankhe28:
image

New change

anth-volk:
Thanks, this looks good! I'll check it out today or tomorrow.

anth-volk (approved):
Thanks for this @psankhe28! This all looks great.

PR #1707, opened 2024-05-03
Removed CodeBlockOld code, and prism-react-renderer dependency

### LINKED ISSUE #1678: Remove deprecated `CodeBlock` code and properly reference files
At present, `CodeBlock` maintains a deprecated block, while using a less-than-ideal system of importing, then re-exporting, the block used in the API documentation pages. We should clean up this file by doing the following:

1. Moving the API's `CodeBlock` to `CodeBlock.jsx`
2. Getting rid of the old block
3. Re-importing the block into the API documentation page
4. Ensuring that we don't use the old block anywhere
5. The old block depends on a package called "Highlight"; if we don't use that, either, we should uninstall it from the app

anmolchhabra21:
## Description

Fixes #1678 
## Changes

Removed the old code, Updated the imports and removed the prism-react-renderer dependency; it was not used/referenced anywhere in the code, and the website was working fine.

## Screenshots

image
## Tests

17/17 Passed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,6 @@
         "moment": "^2.30.1",
         "openai": "^4.40.0",
         "plotly.js": "^2.32.0",
-        "prism-react-renderer": "^2.3.1",
         "react": "^18.3.1",
         "react-bootstrap": "^2.10.2",
         "react-detect-print": "^0.1.2",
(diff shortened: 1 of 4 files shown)


anth-volk (approved):
Thanks for this @anmolchhabra21, we appreciate yet another solid fix. This is approved, but will be merged after some other, smaller changes to the same page to diminish the risk of merge conflicts.

Issue #1486, opened 2024-03-19
About page should use `useCountryId` instead of explicitly popping URL param at index 1


anth-volk:
In `src/redesign/components/About.jsx`, this line:
```
  const countryId = pathParts[1]; // Assumes the countryId is always the second segment in the path
```
Should use `useCountryId`, which is far more robust with error handling.

psankhe28:
I've made the required changes and created the PR https://github.com/PolicyEngine/policyengine-app/pull/1709
Please have a look at it

anth-volk:
Thanks for this, as well, @psankhe28, will review today.

SEEDART007:
@MaxGhenis hey kindly close this issue the pr is already merged

anth-volk:
@SEEDART007 if you look at the code, either due to reversion or error, the `About` page still uses the original URL param popping behavior.

MaxGhenis:
Thank you for filing this issue. Closing as part of the `policyengine-app` (v1) wind-down: policyengine.org is now served by [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2), and this repository is being archived, so this issue is frozen against a codebase that no longer deploys anywhere.

If this is still relevant on the current site, please re-file it against [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2) — we'd welcome it there. Apologies that this sat open through the migration.

PR #1709, opened 2024-05-03
About page

psankhe28:
## Description

#1486 

## Changes
Made use of the hook `useCountryId`

## Screenshots

image

## Tests

17/17 Passed.


anth-volk (approved):
Thanks so much for this, @psankhe28! This all looks good and works well.

PR #1713, opened 2024-05-06
Improve policy reproducibility block

nikhilwoodruff:
## Description

Migrates the policy reproducibility block to using the dictionary input method.

## Changes

As above.

## Screenshots

image

## Tests

No tests added.

### DIFF
--- a/src/data/reformDefinitionCode.js
+++ b/src/data/reformDefinitionCode.js
@@ -17,8 +17,8 @@ export function getReproducibilityCodeBlock(
 
   return [
     ...getHeaderCode(type, metadata, policy),
-    ...getBaselineCode(type, policy, region),
-    ...getReformCode(type, policy, region),
+    ...getBaselineCode(policy, metadata),
+    ...getReformCode(policy, metadata),
     ...getSituationCode(
       type,
       metadata,
@@ -27,7 +27,7 @@ export function getReproducibilityCodeBlock(
       householdInput,
       earningVariation,
     ),
-    ...getImplementationCode(type, region, year),
+    ...getImplementationCode(type, region, year, policy),
   ];
 }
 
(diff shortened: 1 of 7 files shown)


anth-volk:
Just give me the word and I'll work on these tests, they fail because they expect the old format of this component, which had more conditional outputs, and thus more tests.

nikhilwoodruff:
Ah yes please- thanks!

anth-volk:
Could you rebase? I think that should fix the issue.

src/data/reformDefinitionCode.js
```diff
@@ -43,89 +41,36 @@ export function getHeaderCode(type, metadata, policy) {
...
-  // For US reforms, when calculated society-wide, add reported state income tax
-  if (type === "policy" && US_REGIONS.includes(region)) {
-    lines.push("        self.modify_parameters(use_reported_state_income_tax)");
-  }
-
+  let json_str = JSON.stringify(policy.reform.data, null, 2);
+  let lines = [""].concat(json_str.split("\n"));
+  lines[1] = "reform = Reform.from_dict({" + lines[0];
+  lines[lines.length - 1] =
+    lines[lines.length - 1] + ', country_id="' + metadata.countryId + '")';
```
anth-volk, line 73: Does this piece of code need to reformat policies with numbers as keys (e.g., the US income tax brackets)? At the moment, this doesn't sanitize, causing one of the test failures we're seeing, but I'm not sure if it needs to.

nikhilwoodruff: We don't need to- could you remove the test?

anth-volk: Can do

anth-volk (approved):

Issue #1733, opened 2024-05-08
`App.css` contains a (seemingly unused) style containing the character ͼ


anth-volk:
This seems like a pretty strange piece of code. We should confirm that this isn't in use, and if it's not, it should be removed.

anth-volk:
Closed by #1749.

PR #1749, opened 2024-05-09
Changed the style

psankhe28:
## Description

#1733

## Changes

I have removed the styles in App.css that were not used.

## Tests

17/17 tests passed

### DIFF
--- a/src/style/App.css
+++ b/src/style/App.css
@@ -126,10 +126,6 @@ b {
   font-weight: 600;
 }
 
-.ͼ4g {
-  font-weight: 400;
-}
-
 h1 {
   margin-bottom: 25px;
 }


anth-volk:
Thanks! I will review, just in case I somehow missed where this exists.

anth-volk:
Since we opened this issue, I ran into a case where, believe it or not, this weird style actually exists. I'm going to check just to make sure that we do indeed want to remove it, but I think we should be fine.

anth-volk (approved):
Yeah, I think it's better without this. Thanks @psankhe28!

anth-volk:
In the future, when you open a PR, could you put "Fixes #NUMBER," instead of just "#NUMBER"? This allows GitHub to auto-close the issue. Thanks again!

Issue #823, opened 2023-10-28
Capture research filters in URL


MaxGhenis:
Currently the URL only captures the search query, e.g. https://policyengine.org/us/research?search=child+tax+credit

Filtering by topic, location, or author doesn't change the URL.

Use case that brought me here: I wanted to email someone our list of research papers on child tax credits in the US. The above included one post from the UK.

nikhilwoodruff:
Yes though just noting you can already do this manually: https://policyengine.org/us/research?locations=us&search=child+tax+credit

MaxGhenis:
Thank you for filing this issue. Closing as part of the `policyengine-app` (v1) wind-down: policyengine.org is now served by [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2), and this repository is being archived, so this issue is frozen against a codebase that no longer deploys anywhere.

If this is still relevant on the current site, please re-file it against [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2) — we'd welcome it there. Apologies that this sat open through the migration.

PR #1750, opened 2024-05-09
Filtering for author

### LINKED ISSUE #1726: The author selection in the Web App does not seem to work
I don't think it is actually filtering the articles by author:

https://github.com/PolicyEngine/policyengine-app/assets/110687043/72e67fec-6b38-404d-85f7-e5601d1754c3

psankhe28:
## Description

Fixes #1726 

## Changes

I have fixed the filtering of the author

## Tests

17/17 tests passed

### DIFF
--- a/src/pages/Research.jsx
+++ b/src/pages/Research.jsx
@@ -1,3 +1,4 @@
+import { useEffect } from "react";
 import Header from "../layout/Header";
 import Footer from "../layout/Footer";
 import Section from "../layout/Section";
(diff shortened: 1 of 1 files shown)


anth-volk:
Hard to believe that the change could be this simple. I haven't tested yet, but before I do, you have a commit in this that's part of a different issue (#1749). I'm guessing you opened this feature branch (`filter`) off of that branch (`app-css`) instead of off `master`. That's not a problem at all, could you just drop that commit and re-push? Otherwise, this seems like a simple review.

anth-volk:
Also, could you add "Fixes " in front of the issue number? Thanks again.

psankhe28:
Yes, I will do it.

anth-volk (changes requested):
Hi @psankhe28, thanks for your contribution here. Unfortunately, this still doesn't fix the problem. For example, go to http://localhost:3000/us/research, and on the left-hand side, under "Author", click "Max Ghenis". You'll notice that the first article, "PolicyEngine launches state income tax...", was written by Pavel Makarchuk. We're looking for all displayed articles to be by Max Ghenis.

It is possible that the root of the problem is more comprehensive than just something small. This may require some digging to determine the root of.

psankhe28:
> Hi @psankhe28, thanks for your contribution here. Unfortunately, this still doesn't fix the problem. For example, go to http://localhost:3000/us/research, and on the left-hand side, under "Author", click "Max Ghenis". You'll notice that the first article, "PolicyEngine launches state income tax...", was written by Pavel Makarchuk. We're looking for all displayed articles to be by Max Ghenis.
> 
> It is possible that the root of the problem is more comprehensive than just something small. This may require some digging to determine the root of.

I may have missed something, I will look into it. Sorry!

anth-volk (commented):
I'm going to come back to reviewing this either later today or perhaps mid-week. 

We appreciate your contributions. However, this doesn't quite work properly. For example, running this code locally, if you filter for "Max Ghenis", you receive 37 total posts displayed. When adding a country filter, one would expect to either consistently have more articles displayed, or consistently less - essentially, either the filter should look for one of the filtered tags to be present (thus more) or all of them (thus less). However, if I filter for "Max Ghenis" and "United States", I get 11 (fewer), but if I filter for Max and "United Kingdom", I get 44.

I don't think the issue here is your code, I think it's the original implementation of this feature. I'm going to do some thinking, and if you'd still be willing to contribute, I think this might take a more systematic fix that more succinctly addresses the filter function. Also, if possible, we try to avoid using effect hooks, as they can sometimes be far more complex a solution than necessary, introducing not only added complexity, but also more bug potential, so it would be great if we could come up with an implementation that does not require one.

Thanks for your work on this. I'll spend some more time on this later and send my thoughts then.

anth-volk:
Hi @psankhe28, sorry for the delay in getting back to you on this, I just had to do some thinking as to how to actually get this working.

I think this needs a more comprehensive overhaul, unfortunately. I was looking at the original code, and it is just very broken. Here's the technical setup I'm thinking:

* The component would have one “filters” stateful variable, a nested object comprised of three sub-objects: "topics", "locations", and "authors"
  * On load, this would pre-populate the "locations" segment with all items from the country or from the "global" tag (this works within the component currently, but doesn't properly display that they're selected)
  * This would be passed to each respective dropdown selector component
    * If empty, the component would display all options with none selected, otherwise display checks for those listed (this may require overwriting the "value" attribute of the list)
    * On click, each dropdown component would update "filters" by passing it into the setter (ideally via ES6 object spread operator)
  * Via an effect hook, on "filters" change, we would populate respective URL search params (as described in #823)
* Filtering articles
  * A new function would return whether or not an article includes a given tag or a given location or a given author - the current function seems to work incorrectly

This is quite a bit in the way of changes. If you'd like to take this on, we'd be happy to have your contribution. Alternatively, if you'd like something simpler, we can close this out and I can take it on later. Apologies for this being tougher than expected, we just didn't know how broken the existing filters were.

anth-volk:
Closing in light of https://github.com/PolicyEngine/policyengine-app/pull/1953

PR #1754, opened 2024-05-09
Analysis output 'show prompt' block doesn't load

### LINKED ISSUE #1752: Analysis output 'show prompt' block doesn't load
Probably because of the new code block function.

### LINKED ISSUE #1753: Change labor supply chart to be about earnings
@nikhilwoodruff is it still only employment income we're showing the delta of, or also self-employment income? if both let's change from "employment income" to "earnings" here (if not both, let's add it in the package/api)

[Example:](https://policyengine.org/us/policy?focus=policyOutput.laborSupplyImpact&reform=53889&region=enhanced_us&timePeriod=2024&baseline=2)

image

nikhilwoodruff:
Fixes #1752
Fixes #1753

### DIFF
--- a/src/layout/CodeBlock.jsx
+++ b/src/layout/CodeBlock.jsx
@@ -149,8 +149,16 @@ export default function CodeBlock({
         <CodeMirror
           value={data}
           maxHeight={maxHeight ? maxHeight : !isExpanded && "260px"}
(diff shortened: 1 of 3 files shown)


nikhilwoodruff:
Screenshot:

image

PR #1760, opened 2024-05-10
Changed the button size

### LINKED ISSUE #1757: Buttons in household calculator are sized improperly
The video below demonstrates how the calculator is currently functioning. Ideally, the first three buttons (which are almost certainly imported from `/src/controls/Button.jsx`) should have a set width that matches the one inside `src/pages/household/input/VariableEditor.jsx`. These buttons are made to by default sit at 100% width, and so it's likely as simple as copying and pasting the set width value from `VariableEditor` into the three relevant pages with the incorrectly sized button.

https://github.com/PolicyEngine/policyengine-app/assets/14987227/1c4bd9ca-c707-4b24-ac61-1eaf414bfc89

psankhe28:
## Description

Fixes #1757 

## Changes

Made the width of the button consistent.

## Screenshots

https://github.com/PolicyEngine/policyengine-app/assets/84843461/961069a8-a5f7-4ad8-897b-dcb39b9ad181

## Tests

17/17 tests passed.

### DIFF
--- a/src/pages/household/HouseholdIntro.jsx
+++ b/src/pages/household/HouseholdIntro.jsx
@@ -12,7 +12,7 @@ export default function HouseholdIntro() {
       <SearchParamNavButton
         text="Enter my household"
         focus="input.household.taxYear"
-        style={{ marginTop: 20 }}
+        style={{ marginTop: 20, width: 260 }}
       />
     </CenteredMiddleColumn>
   );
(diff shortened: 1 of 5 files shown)


anth-volk:
Could you run `make format` on this?

anth-volk (changes requested):
This all looks great, @psankhe28, but could you just also size the button on the screen below to be consistent? And if the wording ends up not fitting, could you make it wide enough so that they don't wrap? Otherwise, I think this will be good to go. Thanks again!

Screen Shot 2024-05-15 at 11 43 06 PM

PR #1767, opened 2024-05-10
Add utility function in Jest setup that fetches US and UK metadata

### LINKED ISSUE #1450: Add utility function in Jest setup that fetches US and UK metadata
This is used in various places in the tests, so it would be great to have just one fetch at the start, before all tests

czhou578:
## Description

Fixes #1450 

## Changes

Added a utility function in a separate file called ```metadata_fetch.test.js``` that fetches the metadata once in the very beginning before any of the test suites run.

## Tests

No new tests written


czhou578:
@anth-volk, I did a little more research and it seems like the best we can do is to fetch metadata before every single test file. The ```globalSetup``` in jest is just a place to set up configurations or starting a separate server, and not a place to fetch data and then pass that down. The ```setupFilesAfterEnv``` is more tailored to fetching all the metadata we need, but it runs before each test file. What should we do?

[https://stackoverflow.com/questions/66736782/what-is-the-difference-between-globalsetup-setupfiles-setupfilesafterenv-and-b](https://stackoverflow.com/questions/66736782/what-is-the-difference-between-globalsetup-setupfiles-setupfilesafterenv-and-b)

anth-volk:
Ah, I knew I ran into something like this myself when I was working on this a couple weeks ago.

Would you be willing to look into the `make test` recipe and see if it's possible to add a script there to fetch the data and save it as a file that's gitignored, then retool the tests to utilize it? Maybe even refactor the fetch function into one stored in `__setup__`? I think that might be the best way to proceed. I know it seems trivial and like we should just use the multiple fetches, but the metadata is so common throughout the app (the app fetches it immediately on first load, then just keeps passing it down throughout the app) that I think, as we build out these tests, it will be useful to have.

czhou578:
@anth-volk, I wrote a script with a .mjs file extension, since that's the only way I could get it to run from the make file. I keep running into Jest configuration errors when running the test suites. The fetching is working fine, but it seems like Jest is not parsing the .mjs file correctly, even after I specified that specific file type in the config. Do you mind taking a look? 

image

anth-volk:
Interesting. Sure, I'll take a look, though it might not be until tomorrow, depending on priorities. Thanks for your work on this, @czhou578, it's appreciated.

anth-volk:
I checked into the changes you had made, and I think the issue is that a couple of things. First, Jest is still not 100% ES6+-compliant, and we're forced to transpile CommonJS files into ECMAScript modules using Babel. I think that there's a potential this has something to do with the problem.

The other issue, I think, may be that technically the first line isn't running in the context of Jest. I'm not entirely sure on this, but I have a feeling that the value simply can't be passed, and so Jest is somehow trying to run the file and fetch the metadata and failing.

When I had originally written that comment, I wasn't clear enough, I think. I think the best way of doing this might be to use the module you've written to fetch the data, but to actually write it to a file, stored in `src/__tests__/__setup__`, with a `.json` extension. The method would be very similar to what you've done here, but instead of saving the output to the relevant variables, you'd write to file using `fs`, then in the tests themselves, you'd read from that file.

One last critique - there's a couple places in here where you mix pre-ES6 syntax with ES6+. For example, you use both `var` and `const`, but not `let`, and you use `async/await`, but then treat the output as a then-able. While these are valid, we typically prefer ES6+, so `let`/`const` over `var` and, less strictly, `async/await` over `.then()`. I know you were probably just testing out changes, but just in case you weren't aware of this, I wanted to lay it out here.

czhou578:
@anth-volk, I saw this ```document.createRange``` block was recently added in the ```beforeAll``` function. I'm not sure how to integrate this into the metadata fetching script since this doesn't seem to be related. How do I proceed?
image

anth-volk:
Ah, right, thanks for bringing this up. If you could leave that code there, that would be great, as it has nothing to do with data fetching. 

This code was added because CodeMirror uses the `document` object's `range.getBoundingClientRect()` function to somehow assess the size of the screen. However, I forget why, but this function wasn't working properly within Jest's headless Node server, upon which it runs tests. This code replaces that behavior with a Jest mock function that basically does nothing, only returning a couple empty pieces of data, since we're not testing that behavior.

anth-volk:
@czhou578 Thanks, will review!

metadata_fetch.mjs
```diff
@@ -0,0 +1,42 @@
...
+import fs from "fs";
+import path from "path";
+
+// const fetch = require('node-fetch')
+// const fs = require('fs');
+// const path = require('path');
+
+var metadataUS = null;
+var metadataUK = null;
+var metadataCA = null;
```
anth-volk, line 11:
```suggestion
let metadataCA = null;
```

metadata_fetch.mjs
```diff
@@ -0,0 +1,42 @@
+import fetch from "node-fetch";
+import fs from "fs";
+import path from "path";
+
+// const fetch = require('node-fetch')
+// const fs = require('fs');
+// const path = require('path');
+
+var metadataUS = null;
+var metadataUK = null;
```
anth-volk, line 10:
```suggestion
let metadataUK = null;
```

metadata_fetch.mjs
```diff
@@ -0,0 +1,42 @@
+import fetch from "node-fetch";
+import fs from "fs";
+import path from "path";
+
+// const fetch = require('node-fetch')
+// const fs = require('fs');
+// const path = require('path');
+
+var metadataUS = null;
```
anth-volk, line 9:
```suggestion
let metadataUS = null;
```

src/__tests__/__setup__/setup.js
```diff
@@ -1,5 +1,6 @@
 // import a variety of packages
 import { TextEncoder, TextDecoder } from "util";
 
+console.log("does this run once");
```
anth-volk, line 4:
```suggestion
```

src/__tests__/pages/APIDocumentationPage.test.js
```diff
@@ -2,30 +2,29 @@ import APIDocumentationPage, {
...
+  //     `https://api.policyengine.org/${countryId}/metadata`,
+  //   );
+  //   const metadataRaw = await res.json();
+  //   const metadata = metadataRaw.result;
+  //   return metadata;
+  // }
+
+  // metadataUS = await fetchMetadata("us");
+  // metadataUK = await fetchMetadata("uk");
+  // metadataCA = await fetchMetadata("ca");
```
anth-volk, line 28: Could you remove the above commented out block?

anth-volk (changes requested):
This is great, @czhou578, thanks so much for this! I just had a few minor changes to ensure we're in line with ES6+ and remove comments, but otherwise, I think this is good to merge! Thanks again.

czhou578:
Done

anth-volk (approved):
Thank you again for your contributions to this, @czhou578, they are truly appreciated. Approved.

PR #1768, opened 2024-05-11
Added conditional language formatter check

### LINKED ISSUE #1758: `CodeBlock` does not properly handle Python blocks
At present, if a `CodeBlock` component is passed Python, it improperly displays it - check out the first image of the API documentation page below, which is the live app with Python, followed by the second, where its proper display is illustrated. This appears to have arisen within #1743, specifically lines 153-160, where the block was edited to properly show markdown, but then it really originates with the original implementation of the block, which only supported JSON and Python and no markdown.

Ideally, the block should be extensible to whatever language we want - even if we wanted to pass C into the thing, it should properly handle it, assuming that the `language` prop is set in accordance with CodeMirror's syntax. 

Ideally, we should have something that checks if CodeMirror's `langs` contains the formatter that corresponds to the `langauge` prop, and if so, somehow execute it (if this is possible within the constraints of JS), and then fall back to the default of JSON if incorrect.

Screen Shot 2024-05-10 at 6 01 18 PM
Screen Shot 2024-05-10 at 6 03 20 PM

czhou578:
## Description

Fixes #1758 

## Changes

Added a conditional check in ```CodeBlock``` that would render the appropriate language formatter if the language formatter is found, otherwise it will default to the ```.json``` formatter.

## Screenshots

image
image

## Tests

No new tests added.

### DIFF
--- a/src/layout/CodeBlock.jsx
+++ b/src/layout/CodeBlock.jsx
@@ -151,11 +151,7 @@ export default function CodeBlock({
           maxHeight={maxHeight ? maxHeight : !isExpanded && "260px"}
           editable={false}
           extensions={[
-            language === "json"
-              ? langs.json()
-              : language === "markdown"
-                ? langs.markdown()
-                : langs.json(),
+            language in langs ? langs[language]() : langs.json(),
             EditorView.lineWrapping,
           ]}
           theme={espresso}


anth-volk:
Thanks for this, @czhou578! I will review this later today.

anth-volk (approved):
This looks great, @czhou578! I like the approach. Thanks for this contribution.

PR #1783, opened 2024-05-15
Repair errors in Reproduce in Python

### LINKED ISSUE #1782: Issues in `Reproduce in Python` output
Specifically: 

* true should be sanitized to True
* There's an error where "dataset=" is repeated

anth-volk:
## Description

Fixes #1782.

## Changes

Fixes two errors in the Reproduce in Python block: sanitizes outputs to ensure that reform code uses Python `True`, `False`, and `None`; and removes an error that prints `dataset=` twice when using `enhanced_cps` for calculations.

## Screenshots

Screen Shot 2024-05-15 at 11 27 35 PM

## Tests

None have been added.

### DIFF
--- a/src/__tests__/data/reformDefinitionCode.test.js
+++ b/src/__tests__/data/reformDefinitionCode.test.js
@@ -6,6 +6,7 @@ import {
   getReformCode,
   getSituationCode,
   getImplementationCode,
+  sanitizeStringToPython,
 } from "data/reformDefinitionCode";
 import {
   baselinePolicyUS,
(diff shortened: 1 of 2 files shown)


MaxGhenis (changes requested):
Could you move some of this to a helper function to reduce duplication?

nikhilwoodruff (approved):
Agree with @MaxGhenis' comment

anth-volk:
The PR has been updated to refactor the Python sanitization into a utility function and to test said function.

MaxGhenis (approved):

PR #1785, opened 2024-05-16
Add max width to ValueSetter

### LINKED ISSUE #1731: Set max width on input box on ValueSetter component
This is the numerical input box in the image below. Ideally, this should be long enough to reasonably support numbers up to 10 digits (8 plus 2 after decimal), since the same component is used for income input, but not be so long to look bad.

Screen Shot 2024-05-08 at 7 56 28 PM

anth-volk:
## Description

Fixes #1731.

## Changes

This PR removes the previous 100% width setting for the `ValueSetter` subcomponent of the `ParameterEditor` component, replacing it with a max width of 150px, allowing for a reasonable width on most viewpoints, but a guaranteed minimum of 100px at certain breakpoints when a user is setting an "advanced" start and end date, and thus a long-style date is displayed. 150px was chosen to allow for reasonable input of 7-digit-long numbers, such as a high income.

## Screenshots

The below screenshots illustrate the changes alongside the default "onward" date, a standard date period, and an "advanced" date period.

Screen Shot 2024-05-16 at 8 10 48 PM
Screen Shot 2024-05-16 at 8 10 57 PM
Screen Shot 2024-05-16 at 8 11 09 PM

## Tests

None have been added.

### DIFF
--- a/src/pages/policy/input/ParameterEditor.jsx
+++ b/src/pages/policy/input/ParameterEditor.jsx
@@ -356,8 +356,8 @@ function ValueSetter(props) {
     return (
       <StableInputNumber
         style={{
-          width: "100%",
           minWidth: "100px",
+          maxWidth: "150px",
         }}
         key={"input for" + parameter.parameter}
         {...(isCurrency


anth-volk:
@nikhilwoodruff At the largest mobile breakpoints, it looks like the below:

Screen Shot 2024-06-04 at 8 11 43 PM

At the smallest, the input component is clipped off the page, requiring the user to scroll to the right, but this is at about 300-350px, which is smaller than the vast majority of phones
Screen Shot 2024-06-04 at 8 12 03 PM

nikhilwoodruff:
OK that's fine!

PR #1786, opened 2024-05-16
Added button to reset date

### LINKED ISSUE #1736: Create a method to reset the default date setting in the ParameterEditor component
The date component below defaults to "from 2024 onward," but if a user selects a custom setting, there's no way to return to this view. 

Screen Shot 2024-05-08 at 8 30 05 PM

czhou578:
## Description

Fixes #1736 

## Changes

Adds a button next to the value component in ```ParamEditor``` component that will reset the date to the default range.

## Screenshots

image
image

## Tests

No new tests were added.


anth-volk (changes requested):
Thanks for this, @czhou578, this would be a great feature to include. That said, I think it doesn't quite meet the needs of the issue. I believe your changes allow resets if the value sitting within the user policy is set to be a different time period than 2024-01-01 to the pseudo forever date, but what we're looking for is the button to be based upon the state of the date setter. 

I've recorded a video, below, of a sample use case: In this case, a user changes the date for a parameter from "2024 onward" to "2024 to 2029", and at this point, it would be great for a user to be able to reset the date value itself. 

Also, as I think about it, I can see how just calling the button "Reset" could confuse users. Could you change the text to "Reset date" to make the functionality a bit less ambiguous?


https://github.com/PolicyEngine/policyengine-app/assets/14987227/12f1bf84-c12d-4994-b209-7a80b794e2e7

anth-volk:
@czhou578 Actually, I've rethought this a bit. What if instead, we add Ant Design's `UndoOutlined` icon ([link](https://ant.design/components/icon)) to the right of the date when it's not default, and use Ant Design's `Tooltip` to display the text "Reset date" when a user hovers over that button?

czhou578:
I think that sounds good.

czhou578:
Does this fit the requirements?
image

anth-volk:
Is it possible to add it as a "post" tab, akin to what's currently present with the input box to its right, as describe in Ant Design's documentation at https://ant.design/components/input? Even if doing so requires dropping the tooltip and/or somehow "disabling" the icon when the user is in the default "2024 onward" state, instead of conditionally rendering.

Otherwise, yes, I think this is looks great! Thanks again.

anth-volk:
@czhou578 Let me look into this a bit, then, and come up with something. Thanks for your patience.

anth-volk:
@czhou578 In the docs for Ant Design's `Space` component, there are a few examples that mimic what I think would work well here. Would you be willing to look over the docs present at https://ant.design/components/space (especially the part that shows the image I've added below) and see how feasible this is? If not, I can also take a stab at it.

Screen Shot 2024-05-22 at 3 16 46 PM

czhou578:
@anth-volk, thanks for telling me about this component! I believe it should be fixed now. 
image

anth-volk (approved):
@czhou578 This is great, thanks so much for it! Excited to get this merged.

Issue #1300, opened 2024-01-31
Use `react-router` properly


abhcs:
A glance at the [tutorial](https://reactrouter.com/en/main/start/tutorial) tells me that we are underutilizing this package by a mile.

1. The package supports [nested routes](https://reactrouter.com/en/main/start/tutorial#nested-routes) and [handling not found errors](https://reactrouter.com/en/main/start/tutorial#handling-not-found-errors). These features can solve issues such as #1191.
2. The package supports [loading data](https://reactrouter.com/en/main/start/tutorial#loading-data) and [updating data](https://reactrouter.com/en/main/start/tutorial#updating-data). We use `useEffect` blocks for these functionalities, which are linked to issues such as #1176.
3. Some navigation errors such as #923 and #1035 can also be solved by using this package properly.

dotslashbit:
@abhcs I think you are right.
Let me know if you are going to work on it nor not. If not, then I would love to pick this up

abhcs:
I would like to work on it, but it would be great if you could provide feedback on the changes by testing or reviewing the code while it is in a draft state. I expect to open a draft PR next week maybe.

anth-volk:
Since @abhcs filed this, I think it'd only be fair for him to pick it up if interested, but thanks for your enthusiasm @dotslashbit, as Abhishek noted, it'd be great to have you on testing the draft

dotslashbit:
@abhcs thanks, would love to do that

anth-volk:
I spent quite a bit of time working on this Friday and today. I think everything @abhcs has said on this is correct, and with some more advanced features of `react-router`, I was able to get about 70% of things running with a codebase that was much cleaner. The implementation also fixed #1471, #1780, and #1795.

However, there are a couple of major headwinds that proved to be too much (at least for me) to do this at this time. I wanted to document these in case this would help to better break up the tasks necessary to actually solve this issue. The issues I ran into are as below:
1. `react-router` itself, as a package, is known for major version changes that require complete refactors of the entire codebase, especially if relying on more advanced features. This could prove irritating from a long-term package maintenance standpoint.
[...]

Issue #1176, opened 2024-01-12
Enable the eslint rule `react-hooks/exhaustive-deps` in code


abhcs:
On searching the `src` dir for the text `eslint-disable-next-line`, we can see that there are about two dozen errors that have been suppressed by turning off the rule for the next line. 

About 10 of the suppressed errors are `react-hooks/exhaustive-deps` errors. To understand these errors, one must understand the recommended usage of [useEffect](https://react.dev/reference/react/useEffect). The fixes for these errors are expected to be nontrivial: for instance, fixing this error in `VariableEditor` took multiple attempts (https://github.com/PolicyEngine/policyengine-app/commit/66d4eb05a145fa8240ced88579f1c0e3b02b5e03, https://github.com/PolicyEngine/policyengine-app/commit/f13e5d5fc3aad5ef5abc0780a543876624a77c0a, https://github.com/PolicyEngine/policyengine-app/commit/d9fa261d6a366e80ad84e381ce8f4e407c2df899).

Expected: Fix each instance where the rule is suppressed. Enable the rule in each instance by deleting `eslint-disable-next-line`.

Warning: Each fix requires significant testing and [troubleshooting](https://react.dev/reference/react/useEffect#troubleshooting).

MaxGhenis:
Thank you for filing this issue. Closing as part of the `policyengine-app` (v1) wind-down: policyengine.org is now served by [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2), and this repository is being archived, so this issue is frozen against a codebase that no longer deploys anywhere.

If this is still relevant on the current site, please re-file it against [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2) — we'd welcome it there. Apologies that this sat open through the migration.

PR #1803, opened 2024-05-23
Improve use of `react-router`

### LINKED ISSUE #923: When on a calculator page, users need to click the browser back button twice to navigate to their original page

### LINKED ISSUE #1035: Once I choose a calculator I cannot go back using the browser's back button
Once I choose a [calculator](https://policyengine.org/us/calculator), I cannot go back using the browser's back button (or Alt + LeftArrow). I have checked this on Safari and Chrome browsers.

### LINKED ISSUE #1471: Mobile users without geolocation enabled default to `policyengine.org/undefined`, not `/us`
This makes the calculator unusable

### LINKED ISSUE #1740: If API doesn't properly hand back metadata within `PolicyEngine.jsx`, disable login experience and calculators to allow static site to load
The entire app relies upon a metadata fetch at first load. If this fails, we should:
* Disable the login experience (it relies upon the metadata) - This should disable the button, add some sort of disabled visual color and a tooltip saying "Login temporarily disabled"
* Disable the policy and household calculators, as these rely heavily upon metadata - again, the relevant link buttons should have some sort of disabled styling and a tooltip saying "Login temporarily disabled"
* We should somehow obscure/disable the pages if users attempt to access them via URL

### LINKED ISSUE #1780: Uncaught runtime error
I got the following error while running the web app in the localhost.

OS : MacOS 
Device: MacBook Air M2
Browser: Safari

### LINKED ISSUE #1795: Refactor defaultBaselinePolicy code in `PolicyEngine.jsx`
This is a pentuple-nested ternary, which isn't exactly best practice

anth-volk:
## Description

Fixes #1780.
Fixes #1795.
Fixes #923.
Fixes #1035.
Fixes #1740.
Fixes #1471.
This also puts us in a better position to address the following:
#1300 (as it improves our use of `react-router`)
#1176 (as more advanced features from `react-router` may enable diminishing our use of effect hooks)
#1191 (as this could also be done using more advanced features from `react-router`)

## Changes

This PR makes various changes meant to improve our use of `react-router`, thereby removing various routing-induced bugs. Toward this end, this PR does the following:
* Creates a layout component that more stably routes a user who loads our raw URL, `policyengine.org`. Previously, in these cases, we assumed a user's region based on `navigator.language`, but our code made the assumption that their primary browser language code would be "en-", followed by a relevant country, which doesn't apply for users whose primary language is not English or for two-character codes that this method sometimes returns (e.g., just "en"). This would cause the code to return `undefined` as a string, after which the app would treat this as the country ID, failing to load critical pieces of data and crashing. This PR instead uses `Intl.DateTime`, which provides a guaranteed five-character string (e.g., "en-us"). It then pulls only the country ID, ignoring primary language, and if that is not an item within an array of current country versions, the app redirects to "us"
* Creates a country ID routing component that handles more edge cases than previously. For example, in the current app, a user can connect to "policyengine.org/garbage", and "garbage" will be treated as a valid country ID. This routing component will forbid that behavior and redirect to "us".
* Places all routes containing a country ID into a nested routing structure, the current preferred method within `react-router` v6. 
* Tracks errors within the `PolicyEngine.jsx` component's metadata fetch action and conditionally displays an error page for the policy and household calculator pages. This could be refined down the road to only display said error message within the main action portion of the screen.
* Tracks errors within `PolicyEngine.jsx`'s user data fetch, preventing the accessing of the user profile page when this fails. These are typically tied to failures to fetch metadata, but can be triggered separately, hence the separate handling.
* Uses `react-router`'s `replace` feature in select portions of the app to ensure that when we utilize `useSearchParams`, a user can still navigate with the browser's back and forward buttons.

## Screenshots

The first clip below demonstrates changes to the use of the forward and back button, while the second illustrates the impact of the changes when metadata isn't fetched successfully.

https://github.com/PolicyEngine/policyengine-app/assets/14987227/beeb8908-c9c0-4c5b-b55e-ddfcefcc2623

https://github.com/PolicyEngine/policyengine-app/assets/14987227/b782f264-17c4-47a7-8f22-f2704d96e6bb

## Tests

Some of these changes would be subject to testing within a separate PR that's already in review as #1778. Otherwise, no new tests have been added.


nikhilwoodruff (changes requested):
Could we use the PolicyEngine header and footer for the error page?

anth-volk:
@nikhilwoodruff It is now reactivated. The footer is also present, it just begins below this content.

Screen Shot 2024-06-04 at 8 26 47 PM

PR #1816, opened 2024-05-23
Added route for API embed

### LINKED ISSUE #1710: Show API status dashboard from API page
I think we used to have a URL like https://policyengine.org/us/api-status that embedded https://policyengine.betteruptime.com/ but I don't see it now. Let's embed it directly in https://policyengine.org/us/api

czhou578:
## Description

Fixes #1710 

## Changes

Added the route for the API dashboard status embed into the app.

## Screenshots

image

## Tests

No new tests added.

### DIFF
--- a/src/PolicyEngine.jsx
+++ b/src/PolicyEngine.jsx
@@ -31,6 +31,7 @@ import TACPage from "./pages/TermsAndConditions";
 import { useAuth0 } from "@auth0/auth0-react";
 import { ConfigProvider } from "antd";
 import style from "./style";
+import { StatusPage } from "./pages/StatusPage";
 
 const PolicyPage = lazy(() => import("./pages/PolicyPage"));
 const HouseholdPage = lazy(() => import("./pages/HouseholdPage"));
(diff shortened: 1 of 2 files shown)


anth-volk (changes requested):
@czhou578 This is great, thank you very much. I know this wasn't in the initial issue, but would you also be willing to do the following minor thing:

The iframe that the BetterStack display sits in adds that second scrollbar in the screenshot you provided. Could you increase its height from 800px to 1000px?

You'll also notice that the `/calculate` endpoint looks like it fails, even though it actually succeeds. This is because the component uses a hard-coded sample household that is no longer valid. I think that will require deleting particular parts of the default, and it would be hard to describe exactly what should be done, so if you'd be willing, I'll open an issue after the status page is restored, and I'll take that one on myself.

Thanks again!

czhou578:
@anth-volk Done.

anth-volk:
My apologies for losing track of this. Will review.

anth-volk (approved):
Thanks so much for this, @czhou578! Great to have this monitoring available again.

Issue #1809, opened 2024-05-23
Remove deprecated Contact page


anth-volk:
We currently have a Contact page component that is rendered within our routing (accessible at https://www.policyengine.org/us/contact), but that isn't actually finished. The component itself should be deleted and its route should be removed.

anth-volk:
Fixed with #1818.

PR #1818, opened 2024-05-24
Removed deprecated Contact page

memickeymac03:
## Description

Removed the deprecated Contact page as mentioned in #1809 

## Changes

[x] deleted the [Contact.jsx](https://github.com/PolicyEngine/policyengine-app/blob/master/src/pages/Contact.jsx)
[x] removed the dependencies of Contact.jsx (deleted lines 6 and 295 respectively) from [PolicyEngine.jsx](https://github.com/PolicyEngine/policyengine-app/blob/master/src/PolicyEngine.jsx)
[x] removed the path to Contact.jsx effectively removing the contact page


anth-volk:
Thanks for this, @memickeymac03, and welcome to PolicyEngine! We may not be able to review this today, but I will try to do so toward the end of the day. If that isn't possible, don't worry, it will be reviewed, just not until Monday.

anth-volk (approved):
Thanks for this @memickeymac03! This works perfectly.

anth-volk:
In the future, could you write the following in the issue:

Fixes # NUMBER

or 

Closes # NUMBER

Unfortunately, just referencing the number doesn't close it on our end.

PR #1841, opened 2024-05-30
added test for about page

dotslashbit:


src/__tests__/pages/About.test.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import "@testing-library/jest-dom";
+import { screen, render } from "@testing-library/react";
+import About from "../../pages/About";
+import { MemoryRouter } from "react-router-dom";
+
+describe("About Page", () => {
+  test("renders without crashing", () => {
+    render(
+      <MemoryRouter>
```
anth-volk, line 10: This should actually be `BrowserRouter`, as we don't really use `MemoryRouter` in our code (that I know of)

src/__tests__/pages/About.test.js
```diff
@@ -0,0 +1,25 @@
...
+        <About />
+      </MemoryRouter>,
+    );
+
+    const heading = screen.getByRole("heading", { name: /Our people/i });
```
anth-volk, line 15: I like the use of "getByRole" here. Being honest, I didn't even know that was a type of role.

src/__tests__/pages/About.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+    expect(heading).toBeInTheDocument();
+    const max = screen.getByText(
+      "PolicyEngine's team leads a global movement of open-source contributors.",
+    );
+    expect(max).toBeInTheDocument();
+
+    screen.debug(max);
```
anth-volk, line 23: Just to flag: best to remove this line before pushing the test

src/__tests__/pages/About.test.js
dotslashbit: got it

src/__tests__/pages/About.test.js
dotslashbit: Thanks, When I was learning about the query types, it was written that getbyrole is always preferable.

src/__tests__/pages/About.test.js
dotslashbit: got it

anth-volk (approved):
Thanks @dotslashbit, this is perfect! If you're ready to go with this, if you could convert out of draft and run the linter, we can merge.

dotslashbit:
@anth-volk I have added some more tests that we discussed on slack, please take a look, thanks for all the help and support.

anth-volk (approved):
Looking through this, this looks great, @dotslashbit! Thanks so much for your work, it's great to increase our testing.

PR #1846, opened 2024-05-31
test: Research Page

dotslashbit:


src/__tests__/pages/Research.test.js
```diff
@@ -0,0 +1,135 @@
...
+    title: "The Triple Lock Plus",
+    description:
+      "PolicyEngine estimates the impact of the UK government's proposed pension uprating policy reform.",
+    date: "2024-05-28 09:00:00",
+    tags: ["uk", "policy", "featured"],
+    authors: ["nikhil-woodruff"],
+    filename: "uk-triple-lock-plus.ipynb",
+    image: "uk-triple-lock-plus.jpg",
+  },
+];
```
anth-volk, line 69: Could you add one UK post that isn't featured to this list, as well as one that's marked "global" with no country tag? Both of these are cases we might face. You could also probably drop `"PolicyEngine launches state income tax modeling nationwide"`, as it doesn't introduce any new permutation.

anth-volk: Also, if you could add a state-level post (e.g., I think there's a New York one in there) and our sole Nigeria post, I think that would be beneficial.

src/__tests__/pages/Research.test.js
```diff
@@ -0,0 +1,198 @@
...
+    blogPosts.forEach((mockPost, index) => {
+      render(
+        <BrowserRouter key={index}>
+          <MediumBlogPreview key={mockPost.title} blog={mockPost} />
+        </BrowserRouter>,
+      );
+
+      screen.debug();
+
+      const titleElement = screen.getByText(mockPost.title);
```
anth-volk, line 182: Could you do this by using `getByRole("heading", ...)`?

src/__tests__/pages/Research.test.js
```diff
@@ -0,0 +1,198 @@
...
+      expect(titleElement).toBeInTheDocument();
+
+      const descriptionElement = screen.getByText(mockPost.description);
+      expect(descriptionElement).toBeInTheDocument();
+
+      // Check if at least one "Read" link exists
+      const readLinks = screen.getAllByText("Read");
+      const readLink = readLinks.find(
+        (link) =>
+          link.closest(".stretched-link")?.getAttribute("href") ===
```
anth-volk, line 192: This is good, but to avoid breaking if Ant Design ever changes their CSS styles, could you instead use `screen.getAllByRole("link", {name: /READ arrow_forward/i})` to get the links, then use their `value` property to check the href? If you want to view all the properties available to you, if you right-click on an element on our page and click something like "View accessibility properties," that may help to give ideas for details to match against.

anth-volk (approved):
Thanks so much for this, @dotslashbit! This works very well and appears to be very future-proof, as well.

PR #1852, opened 2024-06-06
test: Parameter Editor

dotslashbit:


src/__tests__/pages/policy/input/ParameterEditor.test.js
```diff
@@ -0,0 +1,61 @@
+// src/components/ParameterEditor.test.jsx
+
+import React from "react";
+import { render } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import ParameterEditor from "../../../../pages/policy/input/ParameterEditor";
+import { BrowserRouter } from "react-router-dom";
+
+global.URL.createObjectURL = jest.fn();
```
anth-volk, line 9: Could you elaborate on the purpose of this line?

anth-volk: Would also be great to have a comment in the test saying what its purpose is

src/__tests__/pages/policy/input/ParameterEditor.test.js
```diff
@@ -0,0 +1,61 @@
...
+      description: "This is an example parameter.",
+      period: "monthly",
+      values: {
+        "2024-01-01": 10,
+        "2024-02-01": 20,
+      },
+      unit: "bool",
+    },
+  },
+  countryId: "exampleCountryId",
```
anth-volk, line 31: Could you actually set this one to either "us" or "uk"? That makes it a bit more similar to the actual functioning.

src/__tests__/pages/policy/input/ParameterEditor.test.js
dotslashbit: Understood, at first due to some reason I was facing an error saying createObjectURL is undefined or something like that, so I mocked it and it passed. But now there's no error even after removing this line, so I removed it in the current changes. I have also added comments in the test code

src/__tests__/pages/policy/input/ParameterEditor.test.js
dotslashbit: In the new commit I have added the countryId as 'us'.

src/__tests__/pages/policy/input/ParameterEditor.test.js
```diff
@@ -0,0 +1,52 @@
+// CenteredMiddleColumn.test.js
```
anth-volk, line 1: Just to be clear, do you plan to separate this out into a new file later? If not, it would be best if you could do so and mirror the folder structure of the repo as a whole.

src/__tests__/pages/policy/input/ParameterEditor.test.js
```diff
@@ -0,0 +1,52 @@
...
+import React from "react";
+import { render } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import CenteredMiddleColumn from "/Users/introvertedbot/Documents/Projects/open-source-work/policyengine-app/src/layout/CenteredMiddleColumn.jsx";
```
anth-volk, line 5: I think your import here isn't quite right. Could you import from within the repo?

src/__tests__/pages/policy/input/ParameterEditor.test.js
```diff
@@ -0,0 +1,52 @@
...
+    const { getByText } = render(
+      <CenteredMiddleColumn
+        title="Test Title"
+        description="Test Description"
+      />,
+    );
+    expect(getByText("Test Description")).toBeInTheDocument();
+  });
+
+  test("renders description label by default when description is provided", () => {
```
anth-volk, line 23: I'm not sure I understand the difference between this test and the one above

dotslashbit: One is testing the actual description and another is testing the "description" text is present or not

anth-volk (commented):
The changes you made to CenteredMiddleColumn all look great, @dotslashbit, thanks for the contributions!

anth-volk (approved):
I realize it's been quite a while, @dotslashbit, but I figured I'd remove the empty file that had been the beginning of another series of tests and merge what was here. Thanks again for your work on this back in the day.

PR #1864, opened 2024-06-11
Move default year "onwards" in parameter editor to the next calendar year

### LINKED ISSUE #1863: Move default year "onwards" in parameter editor to the next calendar year
I think this makes more sense and is a more common use-case- plus it's guaranteed to capture any mid-year parameter changes.

cc @MaxGhenis @anth-volk

nikhilwoodruff:
Fixes #1863

## Description

Added 1 to the default year.

## Changes

As above.

## Tests

None added


anth-volk:
Closing as #1863 is not planned

PR #1892, opened 2024-06-19
Fix scroll bars

### LINKED ISSUE #1480: Correct `scroll` settings on policy calculator page
For users who display their scrollbars, our scroll settings damage the app's visuals. The three scroll issues are circled in the image below:
313720242-4c611192-f05c-455d-861b-fd7253f1234b

For item no. 1, the scroll should be entirely removed, and elements in this segment should just shrink until that's no longer feasible for the UX, then should wrap and push the rest of the bar downward.

For items 2 and 3, the horizontal scroll should be removed. Both already implement text wrapping, so it appears that that's the only necessary change.

anth-volk:
## Description

Fixes #1480.

## Changes

This removes unnecessary scroll options so as to prevent the automatic display of scrollbars on browsers where their display is enabled. The only bar that couldn't be removed was on the policy calculator's reform carousel, where the `scroll-y` can't be removed.

## Screenshots

Screen Shot 2024-06-19 at 7 13 11 PM
Screen Shot 2024-06-19 at 7 13 20 PM

## Tests

N/A

### DIFF
--- a/src/layout/StackedMenu.jsx
+++ b/src/layout/StackedMenu.jsx
@@ -3,7 +3,6 @@ import Menu from "./Menu";
 export default function StackedMenu(props) {
   const { firstTree, secondTree, selected, onSelect } = props;
 
-  // onOutput == focus=policyOutput.?
   const isOnOutput =
     window.location.search.includes("focus=policyOutput") ||
     window.location.search.includes("focus=householdOutput");
@@ -15,7 +14,7 @@ export default function StackedMenu(props) {
   }
 
   return (
-    <div style={{ overflow: "scroll", padding: 20, height: "100%" }}>
+    <div style={{ padding: 20, height: "100%" }}>
       {result}
     </div>
   );
--- a/src/layout/StackedMenu.jsx
+++ b/src/layout/StackedMenu.jsx
@@ -13,9 +13,5 @@ export default function StackedMenu(props) {
     result = <Menu tree={firstTree} selected={selected} onSelect={onSelect} />;
   }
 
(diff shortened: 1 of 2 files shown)


nikhilwoodruff:
The screenshots are before right?

anth-volk:
No, these are after. If a user has decided to display scrollbars, then any item with a scroll container will show these. If the user configures their browser to not automatically display scrollbars, these will not show.

anth-volk:
For the sake of comparison, this is what the site looks like for me, even without these changes (which do remove some superfluous scroll containers) because I have scrollbars disabled by default on my system:

Screen Shot 2024-06-19 at 9 31 39 PM

nikhilwoodruff:
> No, these are after. If a user has decided to display scrollbars, then any item with a scroll container will show these. If the user configures their browser to not automatically display scrollbars, these will not show.

I don't understand- how does this solve the issue

anth-volk:
The issue is that there is no way to remove the empty scroll containers in cases where a component needs to scroll and a user's browser is configured to always show scroll bars. This PR removes unnecessary scroll containers, but if a user has set their browser to always show scroll bars, there's no way to remove the bars that we need, other than to re-engineer the UX to not use scrolling.

anth-volk:
As an update - at least on Mac, external mice cause the scroll containers to always be shown, whereas the trackpad causes them to be hidden by default. This would explain why some users constantly see these, and some never do, despite no system changes.

nikhilwoodruff:
Ok- thanks

PR #1947, opened 2024-08-08
Flatten tree and handle error when focus not found in DOM

### LINKED ISSUE #1814: Mobile view arrows crash when navigating
On the mobile interface (this is also the interface present at breakpoints <= 768px, so standard PCs/laptops with the screen open in half-width), pages that have new addresses in order to enable page nesting break the app. 

For example, budgetary output used to be located at `https://policyengine.org/us/policy?focus=policyOutput.budgetaryImpact...`, but now it's at `https://policyengine.org/us/policy?focus=policyOutput.budgetaryImpact.overall...`. However, the mobile output page wasn't updated to support that, so when a user uses an arrow button to navigate, the app crashes on these pages.

rachel-lawrie:
Fixes #1814.

## Description
  Fixes #1814.

## Changes

1. Added a helper function to flatten the tree in https://github.com/PolicyEngine/policyengine-app/blob/master/src/pages/policy/output/tree.js
2. Added error handling that checks that the new focus is present in https://github.com/PolicyEngine/policyengine-app/blob/master/src/pages/policy/output/ImpactTypes.jsx. If it is not present, then the focus is updated to the next index.
3. Added jest tests to identify if any upstream changes breaks the buttons in the future.
4. Removed ability to click breadcrumbs. Note that none of the breadcrumbs in the MobileBottomMenu have pages associated with them, therefore I disabled the “On-Click” capability as a stop-gap-measure before a more comprehensive solution is put in place that equips to the mobile user to see all policy impact categories and click on them.

## Screenshots
https://www.loom.com/share/e3aed8b80f7d4ffb9cc937e1e14b6a0c?sid=1a75872b-44ff-4870-9035-c4b89dc72a5e
https://www.loom.com/share/f1a89fa2ad704174ad007afa2f14df26?sid=4cb67ee1-4c2c-4e59-bf68-f46e7c50b2fb

## Tests
Tested for both 'household' and 'policy' types by running through the app locally for both UK and US.

I also created Jest tests for both 'household' and 'policy' types for both the US and the UK. All tests passed. Note before I made the changes 8 tests were failing and those are still failing.


anth-volk:
One thing before reviewing: could you add your virtual environment to the `.gitignore`?

MaxGhenis:
Please mark as draft

.gitignore
```diff
@@ -33,3 +33,5 @@ yarn-error.log*
 .env
 /src/__tests__/__setup__/data.json
 .idea
+
+venv/
```
MaxGhenis, line 37: could solve the large diff
```suggestion
venv/*

```

rachel-lawrie: Thank you!

MaxGhenis:
Could you edit the description to use keywords that link the pr to the issue? https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

anth-volk:
@rachel-lawrie Max meant could you keep the same title you had, but in the first line in your PR description, add:

Fixes #1814

anth-volk (commented):
Just went through this, thanks @rachel-lawrie! Here are a few thoughts:

* I think the tree flattening code you utilize works well for determining what the right arrow should lead to. However, if you try accessing a breadcrumb in the middle of the tree, the code crashes. Could you keep the flattening for the arrows, but add a component similar to `FolderPage` (this could even mean just making `FolderPage` more generalizable) for the mobile view only so that users can navigate through?
* This isn't related to the tree flattening, but could you remove the extra `BottomCarousel` component in the picture below (this probably means adding code that checks if the user's on mobile, and if so, doesn't show it) and widen the navigation buttons to be the same width as the other buttons on that menu?

Also, if you think the breadcrumbs should sit elsewhere, such as above the buttons in a row, please feel free to implement.

Thanks so much for your work on this! We really appreciate your enthusiasm and your contributions.

.gitignore
```diff
@@ -33,3 +33,5 @@ yarn-error.log*
 .env
 /src/__tests__/__setup__/data.json
 .idea
+
+venv/*
```
MaxGhenis, line 37: We generally keep an empty newline in all files https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline
```suggestion
venv/*

```

rachel-lawrie:
Update: Arrows now work. I am meeting with Anthony to ensure test plan is comprehensive prior to merge.

src/__tests__/pages/household/output/HouseholdMobileCalculator.test.js
```diff
@@ -0,0 +1,186 @@
+/** This test file is designed to identify if any upstream changes break the next and previous buttons in the mobile view. It tests
+ * to ensure the flattenTree function works correctly, that buttons are present as expected, and that when buttons click the focus
+ * updates as expected.
+ */
+
+import "@testing-library/jest-dom";
+import { render, screen, cleanup, fireEvent } from "@testing-library/react";
```
anth-volk, line 7:
```suggestion
import { render, screen, fireEvent } from "@testing-library/react";
```

src/__tests__/pages/household/output/HouseholdMobileCalculator.test.js
```diff
@@ -0,0 +1,186 @@
...
+                    />
+                </MemoryRouter>
+            );
+  
+            // Check the presence of the buttons
+            const prevButton = screen.queryByTestId('prev-button');
+            const nextButton = screen.queryByTestId('next-button');
+  
+            if (index > 0) {
+              expect(prevButton).toBeInTheDocument();
```
anth-volk, line 77: Throughout this test, the linter returns a warning not to place `expect()` inside a conditional, as there are cases where errors thrown by the code aren't recognized. Could you refactor this, even if it involves writing three separate tests, one for `index[0]`, one for `index[length - 1]`, and one for all others?

anth-volk: Alternatively, if you write a small helper function that just checks the index and determines what should be returned, then `expect`ed the output of that function (something like `expect(sampleFunc(prevButton))`, and then have `sampleFunc(arg)` selectively either return `null` or the original `arg`

src/layout/MobileCalculatorPage.jsx
```diff
@@ -13,6 +20,7 @@ import HOUSEHOLD_OUTPUT_TREE from "../pages/household/output/tree";
 import VariableSearch from "../pages/household/VariableSearch";
 import { ParameterSearch } from "../pages/PolicyPage.jsx";
 import PolicyRightSidebar from "../pages/policy/PolicyRightSidebar.jsx";
+import { impactKeys } from "../pages/policy/output/ImpactTypes.jsx";
```
anth-volk, line 23: I couldn't add the comment at the correct line, but could you remove line 17 (`import {copySearchParams}...`) if it's no longer called?

src/layout/MobileCalculatorPage.jsx
```diff
@@ -330,15 +386,32 @@ function MobileBottomNavButtons({ focus, type, metadata }) {
...
+  // define valid focus values for checking if focus for previous and next is valid
+  const validFocusValues = impactKeys;
+
+  // Define currentIndex after options is determined
   const currentIndex = options.map((option) => option.name).indexOf(focus);
-  const previous = options[currentIndex - 1] || {};
-  const next = options[currentIndex + 1] || {};
+
+  // Apply logic based on type
+  let previous, next;
```
anth-volk, line 406: Could you explicitly instantiate these to `null`?

anth-volk (changes requested):
Thank you very much for this contribution @rachel-lawrie. I just had some pretty brief changes to request to fix some linting concerns. Otherwise, this looks great! Looking forward to getting merged.

anth-volk (changes requested):
There's an error in this that I didn't notice before. For some reason, despite the fact that it's included in `tree.js`, the `codeReproducibility` (Reproduce in Python) component does not show up for me as an option within the arrows. Once the user clicks through to the AI Analysis page, the right arrow is no longer visible, and thus there's no way to actually visit the Reproduce in Python page on mobile.

anth-volk (approved):
Just ran through a review and this all looks great! Thanks again for your contributions here, @rachel-lawrie!

PR #1953, opened 2024-08-09
Changes in logic to fix filters in Research component

### LINKED ISSUE #1726: The author selection in the Web App does not seem to work

federicopoi:
Fixes #1726 

## Description

This update addresses issues with the filters in the research component and enhances code legibility through refactoring and simplification.

## Changes

**Initialization:** States for filterTopics, locations, and authors are now initialized with empty arrays to ensure that the filtering logic starts with no pre-selected options, allowing for a more predictable user experience.

**Filtering Logic:** Extracted and modularized the filtering logic to improve maintainability and make the code more reusable.

**ResearchExplorer Function:** Moved some functions to helper functions to allow them to be called from the top, to make it easier to read. Also updated to reflect changes in filtering logic and state initialization, ensuring compatibility with the refactored components.

**BlogPostResults Function:** Cleaned up and simplified the function for better usability.

**ExpandableCheckBoxList Function:** Improved readability; removed "Select All" (can be re-added if needed).

**Expandable Function:** Cleaned up and simplified the function for better usability.

### DIFF
--- a/src/pages/Research.jsx
+++ b/src/pages/Research.jsx
@@ -55,11 +55,35 @@ function ResearchExplorer() {
   const displayCategory = useDisplayCategory();
   const [searchParams, setSearchParams] = useSearchParams();
 
-  // Initialize filter states with empty arrays
-  const [filteredTopics, setFilteredTopics] = useState([]);
-  const [filteredLocations, setFilteredLocations] = useState([]);
-  const [filteredAuthors, setFilteredAuthors] = useState([]);
+  // Extract the country ID from the URL pathname
+  const extractCountryIdFromPathname = () => {
+    const pathSegments = window.location.pathname.split("/").filter(Boolean);
+    if (pathSegments.length > 0) {
+      return pathSegments[0];
+    }
+  };
 
+  // Determine initial locations based on the country ID
+  const initialLocations = locationTags.filter(
+    (location) =>
+      location === extractCountryIdFromPathname() ||
+      location.startsWith(extractCountryIdFromPathname() + "-") ||
+      location === "global",
+  );
+
+  // Initialize filter states with the appropriate values
+  const [filteredTopics, setFilteredTopics] = useState(
+    searchParams.get("topics")?.split(",") || topicTags,
+  );
+  const [filteredLocations, setFilteredLocations] = useState(
+    searchParams.get("locations")?.split(",") || initialLocations,
+  );
+  const authorKeys = Object.keys(authors);
+  const [filteredAuthors, setFilteredAuthors] = useState(
+    searchParams.get("authors")?.split(",") || authorKeys,
+  );
+
+  // Filter posts based on the selected criteria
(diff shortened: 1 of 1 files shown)


anth-volk (changes requested):
Thank you very much for your contribution on this, @federicopoi. I only had one revision to request: for each country version, we like to pre-filter such that only global articles and articles from the particular country are displayed. Would it be possible to add that as an initial setting? Otherwise, I think this all looks good to go.

federicopoi:
@anth-volk let me know if that works! Federico

src/pages/Research.jsx
```diff
@@ -54,86 +54,49 @@ export default function Research() {
...
   const initialLocations = locationTags.filter(
     (location) =>
       location === extractCountryIdFromPathname() ||
       location.startsWith(extractCountryIdFromPathname() + "-") ||
       location === "global",
   );
 
+  // Initialize filter states with the appropriate values
   const [filteredTopics, setFilteredTopics] = useState(
     searchParams.get("topics")?.split(",") || topicTags,
```
anth-volk, line 76:
```suggestion
    searchParams.get("topics")?.split(",") || [],
```
Here and below, I'm using an empty array because I think this is our desired default - for everything other than locations, it should be no filters selected instead of all selected, even though the outcome is nearly the same for the user.

src/pages/Research.jsx
```diff
@@ -54,86 +54,49 @@ export default function Research() {
...
   );
   const [filteredLocations, setFilteredLocations] = useState(
     searchParams.get("locations")?.split(",") || initialLocations,
   );
-
   const authorKeys = Object.keys(authors);
-
   const [filteredAuthors, setFilteredAuthors] = useState(
     searchParams.get("authors")?.split(",") || authorKeys,
```
anth-volk, line 83:
```suggestion
    searchParams.get("authors")?.split(",") || [],
```

src/pages/Research.jsx
```diff
@@ -54,86 +54,49 @@ export default function Research() {
...
 
+  // Initialize filter states with the appropriate values
   const [filteredTopics, setFilteredTopics] = useState(
     searchParams.get("topics")?.split(",") || topicTags,
   );
   const [filteredLocations, setFilteredLocations] = useState(
     searchParams.get("locations")?.split(",") || initialLocations,
   );
-
   const authorKeys = Object.keys(authors);
```
anth-volk, line 81:
```suggestion
```
With the changes below on line 83, this line would no longer be necessary

anth-volk (changes requested):
Thanks for your work on this @federicopoi! Just another minor change, then I think it'll be mergeable.

anth-volk (approved):

PR #1955, opened 2024-08-10
Handle deprecated user policies

### LINKED ISSUE #1950: Add helpful error message when policies use incompatible parameters
If we find ourselves at a link where parameters are failing (we should check that they are valid), we should:
* Display a dialogue to the user explaining which parameters (or how many) are invalid/outdated
* Give the user a link to a policy with those removed

anth-volk:
## Description

Fixes #1950.

## Changes

Prior to this PR, users with policies where one or more parameters were deprecated within subsequent country package versions (usually through deletion of the parameter) experience an empty white screen or endless spinner when returning to their policy's policy output URL. This PR does the following:

* Updates the parts of the code dependent upon the `policy` object such that the `policyOutput` page doesn't crash when deprecated parameters are present
* Creates a new `DeprecationModal` that informs users that one or more parameters are deprecated
* Allows the user to transfer the non-deprecated parameters to a new policy

## Screenshots

The below is tested with the deprecated Wyden-Smith policy available at https://policyengine.org/us/policy?focus=policyOutput.policyBreakdown&reform=44066&region=us&timePeriod=2024&baseline=2

https://github.com/user-attachments/assets/8ef074fe-6a93-49f4-84b7-01d0b8bd4e6b

## Tests

Tests have been added. Some remain under development, hence the draft status of this PR, but the logic underpinning the code additions is complete.


src/modals/DeprecationModal.jsx
```diff
@@ -0,0 +1,166 @@
...
+          fontWeight: 500,
+          fontSize: 20,
+        }}
+      >
+        Your policy is deprecated
+      </h6>
+      <p>
+        {`Unfortunately, as of policyengine-${countryId} v.${countryVersion}, some of your parameters are no longer supported:`}
+      </p>
+      {deprecatedParamsJSX}
```
nikhilwoodruff, line 107: Could we use code-block styling for the parameters and replace the main text with label?

nikhilwoodruff (approved):

src/modals/DeprecationModal.jsx
anth-volk: We don't have access to the parameter's label at this point, as the entire item does not exist in the metadata any longer. Additionally, if we did, in some cases, the label isn't very descriptive (e.g., "7" for certain tax brackets).

anth-volk: This is how this now looks.

Screen Shot 2024-08-12 at 7 07 32 PM

anth-volk: Following a minor style edit, here's how it would look with 2+ params (the duplicate is only for visual testing):

Screen Shot 2024-08-12 at 7 10 58 PM

nikhilwoodruff: > We don't have access to the label

Oh of course lol.

Looks good- can we shade in the background like `this` though?

anth-volk: This is now the look with one and with more than one param:

Screen Shot 2024-08-13 at 1 48 57 PM
Screen Shot 2024-08-13 at 1 49 12 PM

MaxGhenis:
Turns out this will be very helpful for a project @PavelMakarchuk and I are working on right now, so I'm going to merge it. Let's address lingering issues in future PRs. Thanks!

PR #1994, opened 2024-08-24
Fix: Make Table of Contents scroll to relevant section of Blog posts

### LINKED ISSUE #1990: Contents of blog post do not lead to the actual section of the blog post
It creates the impression that those are clickable but they are currently not

tawandamoyo:
Fixes #1990 

## Changes

- fix `markdown-formatter` to assign unique and correct heading ids 
- scroll to relevant section of blog post from table of contents

### DIFF
--- a/src/layout/MarkdownFormatter.jsx
+++ b/src/layout/MarkdownFormatter.jsx
@@ -2,7 +2,7 @@ import Markdown from "react-markdown";
 import { TwitterTweetEmbed } from "react-twitter-embed";
 import rehypeRaw from "rehype-raw";
 import remarkGfm from "remark-gfm";
-import { useState, useRef, useEffect } from "react";
+import React, { useState, useRef, useEffect } from "react";
 import style from "../style";
 import useDisplayCategory from "../hooks/useDisplayCategory";
 import Plot from "react-plotly.js";
@@ -370,17 +370,17 @@ export function MarkdownFormatter({ markdown, backgroundColor, dict }) {
           }
         },
         h2: ({ children }) => {
-          let headerText = children[0];
+          let headerText = React.Children.toArray(children).join("");
           if (!headerText.split) {
             headerText = "";
           }
           // Remove slashes and commas, and replace spaces with dashes to create a
           // unique ID for each header.
           const slug = headerText
-            .split(" ")
-            .join("-")
-            .replace("/", "")
-            .replace(/,/g, "");
+            .toLowerCase()
+            .replace(/[/,]/g, "")
+            .replace(/\s+/g, "-");
+
           return (
             <h2 id={slug} style={{ marginBottom: 20 }}>
               {children}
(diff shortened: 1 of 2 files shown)


anth-volk:
Before reviewing, would you mind giving this PR a more descriptive name? No need to reference issue number or anything. Thanks Tawanda!

tawandamoyo:
> Before reviewing, would you mind giving this PR a more descriptive name? No need to reference issue number or anything. Thanks Tawanda!

Sure, done. Thanks

src/layout/MarkdownFormatter.jsx
```diff
@@ -370,38 +370,40 @@ export function MarkdownFormatter({ markdown, backgroundColor, dict }) {
           }
         },
         h2: ({ children }) => {
-          let headerText = children[0];
+          let headerText = React.Children.toArray(children).join("");
```
anth-volk, line 373: Thanks for your work on this, Tawanda. While this may work, there are a lot of reasons why you probably wouldn't want to use `React.Children` in this code; you can read more about why at https://react.dev/reference/react/Children. In a nutshell, it can make the code very fragile, so we would ask that you replace this implementation with a different one.

anth-volk: In fact, a great reason why `React.Children` could be problematic here is if we intend to use `MarkdownFormatter` in a broader way, to format any markdown we might want to display. Doing so would keep our styling consistent and make it easier to build out new markdown-driven pages.

anth-volk, line 373:
```suggestion
```

As I was reviewing this in-depth, I think I found an easier fix than `React.Children` and all that. Here's what's interesting: whoever wrote this before expected `children` to be an array. That's why they wrote:
```
let headerText = children[0]
```
However, if you run this locally and console.log `children`, it's just a string. The easiest way to fix this is to rename `children` to `headerText` (as I do in the comments above and below) and just remove this line. 


Incidentally, this code worked in the past:
```
let headerText = children[0]
```
because a `String` type is really just an array of individual characters. Let's say `children` equaled `My Title`. `headerText` would then equal `M`. It works in JavaScript, but obviously was a bug.

src/layout/MarkdownFormatter.jsx
```diff
@@ -370,38 +370,40 @@ export function MarkdownFormatter({ markdown, backgroundColor, dict }) {
           }
         },
         h2: ({ children }) => {
```
anth-volk, line 372:
```suggestion
        h2: ({ children: headerText }) => {
```

src/layout/MarkdownFormatter.jsx
```diff
@@ -370,38 +370,40 @@ export function MarkdownFormatter({ markdown, backgroundColor, dict }) {
...
+            .replace(/[/,]/g, "")
+            .replace(/\s+/g, "-");
+
           return (
             <h2 id={slug} style={{ marginBottom: 20 }}>
               {children}
             </h2>
           );
         },
         h3: ({ children }) => {
```
anth-volk, line 390:
```suggestion
        h3: ({ children: headerText }) => {
```

src/layout/MarkdownFormatter.jsx
```diff
@@ -370,38 +370,40 @@ export function MarkdownFormatter({ markdown, backgroundColor, dict }) {
...
-          const headerText = children[0];
-          return (
-            <h3 id={headerText.split(" ").join("-").replace(/,/g, "")}>
-              {children}
-            </h3>
-          );
+          const headerText = React.Children.toArray(children).join("");
```
anth-volk, line 391:
```suggestion
```

src/layout/MarkdownFormatter.jsx
```diff
@@ -370,38 +370,40 @@ export function MarkdownFormatter({ markdown, backgroundColor, dict }) {
...
+          const slug = headerText
+            .toLowerCase()
+            .replace(/[/,]/g, "")
+            .replace(/\s+/g, "-");
+
+          return <h3 id={slug}>{children}</h3>;
         },
         h4: ({ children }) => {
```
anth-volk, line 399:
```suggestion
        h4: ({ children: headerText }) => {
```

src/layout/MarkdownFormatter.jsx
```diff
@@ -370,38 +370,40 @@ export function MarkdownFormatter({ markdown, backgroundColor, dict }) {
...
-          const headerText = children[0];
-          return (
-            <h4 id={headerText.split(" ").join("-").replace(/,/g, "")}>
-              {children}
-            </h4>
-          );
+          const headerText = React.Children.toArray(children).join("");
```
anth-volk, line 400:
```suggestion
```

anth-volk (changes requested):
Thanks for your contributions on this issue, @tawandamoyo. I had a couple comments I think you might find valuable to your learning journey in React, and also suggest a different approach that's less involved.

src/layout/MarkdownFormatter.jsx
tawandamoyo: Thanks, have corrected

anth-volk (approved):
Confirmed that this works! Thanks so much @tawandamoyo

Issue #1975, opened 2024-08-16
Bug in ARPA EITC post footnote


MaxGhenis:
https://policyengine.org/us/research/restoring-arpa-eitc

<img width="887" alt="image" src="https://github.com/user-attachments/assets/92516603-7ad9-436d-9aa2-37fc55fab190">

PR #1995, opened 2024-08-29
Fix: Displaying unwanted number instead of required list value; Bug in ARPA EITC post footnote

bharateshwq:
## Description
#1975
The issue with the value extraction logic has been addressed.
## Changes

Previously, the code split the link string at the "-" character and assigned the portion after it to the value. I have introduced additional validation to ensure that only numeric values are accepted for value in all relevant scenarios, as required by the original implementation.


MaxGhenis:
could you drop a screenshot or video?

bharateshwq:
Screenshot 2024-09-01 at 09-48-05 PolicyEngine

PR #1999, opened 2024-09-01
fixed formatting of footnotes in markdowns

### LINKED ISSUE #1975: Bug in ARPA EITC post footnote

bharateshwq:
## Description
Fixes #1975
The issue with the value extraction logic has been addressed.
## Changes

Previously, the code split the link string at the "-" character and assigned the portion after it to the value. I have introduced additional validation to ensure that only numeric values are accepted for value in all relevant scenarios, as required by the original implementation.

Screenshot 2024-09-01 at 09-48-05 PolicyEngine


anth-volk:
Here, too, could you write "Fixes #1975"?

anth-volk (approved):
Thanks @bharateshwq! This works wonderfully.

PR #2001, opened 2024-09-02
Fix terms of service page to say that we follow the selected country laws. Defaults to US law if invalid country id is used

### LINKED ISSUE #1849: UK Terms of Service page describes us as following US law
If this is inaccurate (I'm assuming it is for the UK app), we should update this to say UK law. 

Screen Shot 2024-06-03 at 4 23 32 PM

tawandamoyo:
## Description

Fixes #1849.

## Changes

Fix the terms of service page to specify that we follow the laws of the relevant country, instead of US law

### DIFF
--- a/src/pages/TermsAndConditions.jsx
+++ b/src/pages/TermsAndConditions.jsx
@@ -4,17 +4,24 @@ import { Helmet } from "react-helmet";
 import Header from "../layout/Header";
 import Footer from "../layout/Footer";
 import useCountryId from "../hooks/useCountryId";
+import { COUNTRY_NAMES } from "../data/countries";
 
 export default function TACPage() {
   const mobile = useMobile();
 
-  const countries = {
-    ca: "Canada",
-    uk: "the United Kingdom",
-    us: "the United States",
-  };
   const countryId = useCountryId();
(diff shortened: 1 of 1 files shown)


src/pages/TermsAndConditions.jsx
```diff
@@ -3,9 +3,19 @@ import useMobile from "../layout/Responsive";
 import { Helmet } from "react-helmet";
 import Header from "../layout/Header";
 import Footer from "../layout/Footer";
+import useCountryId from "../hooks/useCountryId";
 
 export default function TACPage() {
   const mobile = useMobile();
+
+  const countries = {
```
anth-volk, line 11: Very minor, but to ensure consistency (and avoid redefining), could you instead use the `COUNTRY_NAMES` object from https://github.com/PolicyEngine/policyengine-app/blob/master/src/data/countries.js?

anth-volk (changes requested):
Thanks, @tawandamoyo! This is a good implementation, but I just had one minor ask to improve consistency.

anth-volk (approved):
Thanks so much @tawandamoyo! This works as intended. Great to get this fixed.

Issue #1735, opened 2024-05-08
Dates in the research cards on the Research page and in the article sections on the homepage should be internationalized


anth-volk:
As the below image shows, at the moment, the UK cards (both on the homepage and the Research page) utilize a US date. I belive there's already a date internationalization function in the repo, probably in the lang folder.

<img width="951" alt="Screen Shot 2024-05-08 at 8 28 13 PM" src="https://github.com/PolicyEngine/policyengine-app/assets/14987227/340f4ee3-8f83-4ea8-b2ec-863a34316578">

PR #2011, opened 2024-09-09
Internationalize dates on articles

tawandamoyo:
Fixes: #1735 

- Use date internationalization function to show internationalized dates on research articles and featured blog posts


anth-volk (approved):
Thanks for this @tawandamoyo! Especially appreciate you being thorough with checking where dates were printed incorrectly.

PR #2013, opened 2024-09-16
Remove the `map_to = person` code from the reproduce in python output

### LINKED ISSUE #2012: Remove the `map_to = person` code from the reproduce in python output
Since we are computing `household_net_income`, mapping it to each person by default could create confusion as it will by return inaccurate results - given the [entity mismatch](https://github.com/PolicyEngine/policyengine-us/blob/9c723ae538c11252ee539ae259856b4ec484415a/policyengine_us/variables/household/income/household/household_net_income.py#L6)

PavelMakarchuk:
Fixes #2012


anth-volk (approved):
Looks great! Thanks @PavelMakarchuk

PR #2017, opened 2024-09-17
Changing from enhanced_cps_2022 to 2024

### LINKED ISSUE #2016: Create default dataset constant for US package and replace all instances of `enhanced_cps_2022` with it
The default value should now be `enhanced_cps_2024`. Areas where we know we use this value include the Reproduce in Python on the policy calculator (still uses 2022).

bharateshwq:
Fixes  #2016
## Description
Changed required files

### DIFF
--- a/src/data/countries.js
+++ b/src/data/countries.js
@@ -54,11 +54,11 @@ export const STATUS_TEXT_COLORS = {
 };
 
 // Map region entries to default datasets;
-// at the moment, this only applies to the 
+// at the moment, this only applies to the
 // "enhanced_us" region selection
 export const DEFAULT_DATASETS = {
-  enhanced_us: "enhanced_cps_2024"
-}
+  enhanced_us: "enhanced_cps_2024",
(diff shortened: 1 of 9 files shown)


anth-volk:
Thank you for this, @bharateshwq. While this does change the underlying specifications in some places, I'm hoping we could create a JS object (probably in something like `/api/countries.py`) that specifies each country's default dataset in the following format:

```
const VAR_NAME: {
  us: "enhanced_cps_2024",
  uk: ...
  ...
};
```

Then we could import that where relevant. If a dataset is only specified in cases of the US, then there would ideally be some sort of checking done, instead. Then, we'd drop this in everywhere we specify a dataset. This way, we only have to change one value in the future, when let's say `enhanced_cps_2026` comes out.

anth-volk (approved):
Thanks for your work on this @bharateshwq. Just wanted to get this over the line, so I made a few small edits at the end.

PR #2021, opened 2024-09-17
Extra quotation marks in the reproduce in python code

### LINKED ISSUE #2020: Extra quotation marks in the reproduce in python code
[Here](https://github.com/PolicyEngine/policyengine-app/blob/master/src/data/reformDefinitionCode.js#L181)

PavelMakarchuk:
Fixes #2020


anth-volk (approved):
Good catch, thanks for fixing @PavelMakarchuk

PR #2028, opened 2024-09-24
Add option for 'lite' simulations

### LINKED ISSUE #2027: Add option for 'lite' simulations
By default, we could run simulations on a smaller sample of each dataset.

nikhilwoodruff:
Fixes #2027

Open to feedback or rejection.

https://github.com/user-attachments/assets/b802286c-2a1e-41ca-b474-e25e6d907c52

### DIFF
--- a/src/pages/policy/output/FetchAndDisplayImpact.jsx
+++ b/src/pages/policy/output/FetchAndDisplayImpact.jsx
@@ -36,7 +36,7 @@ export function FetchAndDisplayImpact(props) {
   const timePeriod = searchParams.get("timePeriod");
   const reformPolicyId = searchParams.get("reform");
   const baselinePolicyId = searchParams.get("baseline");
-  const maxHouseholds = searchParams.get("mode") == "lite" ? 10_000 : null;
+  const maxHouseholds = searchParams.get("mode") == "full" ? null : 10_000;
   const renamed = searchParams.get("renamed");
 
   const [impact, setImpact] = useState(null);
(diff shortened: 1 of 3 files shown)


anth-volk:
Would you mind re-uploading the video? Doesn't display for me.

anth-volk:
Also, there's a bug whereby you used loose equality (`==`) instead of strict (`===`)

src/pages/policy/output/FetchAndDisplayImpact.jsx
```diff
@@ -36,6 +36,7 @@ export function FetchAndDisplayImpact(props) {
   const timePeriod = searchParams.get("timePeriod");
   const reformPolicyId = searchParams.get("reform");
   const baselinePolicyId = searchParams.get("baseline");
+  const maxHouseholds = searchParams.get("mode") == "full" ? null : 10_000;
```
anth-volk, line 39:
```suggestion
  const maxHouseholds = searchParams.get("mode") === "full" ? null : 10_000;
```

anth-volk:
I think the "lite" simulations option looks great, but is the timing component working correctly? It doesn't appear in the video, but perhaps it was added after that was taken.

anth-volk:
Curious to hear Max's thoughts

MaxGhenis:
Can we start by making lite the non-default option? Including rephrasing the toggle as such. In general looking to avoid changing default behaviors unless necessary--we should give users some sort of notice analogous to deprecation warnings.

anth-volk:
@MaxGhenis Correct me if I'm wrong, Nikhil, but my understanding on the states is that their datasets are too small to be affected by the toggle.

MaxGhenis (approved):

PR #2041, opened 2024-09-28
solves undefined variable employment_income error in Marginal tax Rates

### LINKED ISSUE #2039: Page Crashes when clicked on marginal tax rates
Below video demonstrates the issue

https://github.com/user-attachments/assets/0a5bc7cf-ae4b-4721-adce-c93f105a2eca

householdData.people.you.employment_income is undefined
./src/pages/household/output/MarginalTaxRates.jsx/MarginalTaxRates/<@http://localhost:3000/static/js/src_pages_HouseholdPage_jsx.chunk.js:4373:5
commitHookEffectListMount@http://localhost:3000/static/js/bundle.js:391313:30
commitPassiveMountOnFiber@http://localhost:3000/static/js/bundle.js:392806:42

Could reproduce this in chrome and firefox

bharateshwq:
Fixes #2039

solution enables existing code to work as expected even if parts of the object are missing

### DIFF
--- a/src/api/variables.js
+++ b/src/api/variables.js
@@ -356,7 +356,7 @@ export function getValueFromHousehold(
   }
   if (!timePeriodValues) {
     console.error("Error getting variable value", variable);
-    return null;
+    return 0;
   }
   if (!timePeriod) {
     const possibleTimePeriods = Object.keys(timePeriodValues);
--- a/src/pages/household/output/MarginalTaxRates.jsx
+++ b/src/pages/household/output/MarginalTaxRates.jsx
@@ -51,6 +51,7 @@ export default function MarginalTaxRates(props) {
     householdInput,
     metadata,
   );
+
   const currentMtr = getValueFromHousehold(
     "marginal_tax_rate",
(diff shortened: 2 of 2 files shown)


bharateshwq:
Hey may i know why optional chaining has been disabled in the codebase, just curious

anth-volk:
@bharateshwq Could you give more detail on what you mean about optional chaining being disabled?

anth-volk:
@bharateshwq Are you referring to the limited use of `?.` in JS? If so, that has merely been the preference of particular devs in the past. We don't have a hard and fast rule, and in my opinion, there's value to using and not using. Are you saying we have a rule somewhere against it?

bharateshwq:
Hey 
Thank you for your response.
Yes I meant ?. Oh I see that makes sense actually. I kinda got a eslint warning regarding it and then couldn't recreate the error again.
But thanks it's always good to know

anth-volk (approved):
This does indeed fix the problem. Thanks @bharateshwq!

PR #2103, opened 2024-10-17
Infinity inputs

### LINKED ISSUE #2095: Add the ability to set `Infinity` and `-Infinity` as input values
This will require some thought. I think we'll want to restrict this to certain data types (obviously, doesn't make sense for percents, for example), and I'm not sure yet how to integrate into the UI. Opening a separate issue at /PLACEHOLDER/ for enabling in API.

anth-volk:
## Description

Fixes #2095

## Changes

https://github.com/PolicyEngine/policyengine-api/pull/1887 must be merged before this.

This PR makes a few relevant changes to enable infinity and -infinity inputs:
* Adds JSON replacer and reviver functions to JSON.parse() and JSON.stringify(), then builds wrappers on top of these and uses them wherever we anticipate possible receiving an infinite value (anything policy-related)
* Adds a wrapper on top of this for `Response.json()`, which has no means of specifying a custom reviver, and replaces all uses of `Response.json()` with the new wrapper
* Replaces `StableInputNumber` in the `ParameterEditor` component with Ant Design's `InputNumber`, then makes the input a controlled input to allow custom updating. This also required some changes to the underlying display logic.
* Adds a small custom input toolbar present for all inputs except Booleans, percentages, and Enums. This bar has a button to open, and when opened, displays an infinity and negative infinity symbol.

There remain four unaddressed issues that I've chosen to leave unaddressed to make sure we can push this as quickly as possible. If pushed, I can address these at a different time:
1. The parameter history chart does not display infinite values. This is due to limitations within Plotly, and I've yet to find a good solution.
2. No tests have been added, but I intend to do so down the road, considering the priority level.
3. The "Reproduce in Python" page does not properly set relevant reform values to infinity. This shouldn't be too significant.
4. There's a visual bug in the number input box when a parameter has a default value of `Infinity`; for some reason, instead of displaying the infinity symbol, it will display "Infinity," and there are some quirks around that, as well.

I have also tested manually that this does generate a response from the back end when running locally with the required API changes. That said, I had no value to test against, so am unclear on the accuracy of the output.

## Screenshots

https://github.com/user-attachments/assets/af93c58d-ab99-4961-b9d4-7033e63d2960

## Tests

None have been added.


nikhilwoodruff (approved):
Good to merge after adding shading to the open/collapse button, and passing tests.

PR #2116, opened 2024-10-21
Add multi-year scale input method

### LINKED ISSUE #2067: Parameter table mode
I think this could save a lot of time for people programming reform details of complex reforms. Instead of showing the date picker and the chart, we could show an editable table for 2024-2034 (inclusive) in both the UK and US, along with a save button. We could make this non-default, with the old layout shown until the user clicks a switch.

I'd strongly push for this even as an initially experimental feature hidden behind a URL parameter.

cc @MaxGhenis @anth-volk

anth-volk:
## Description

Fixes #2067.

## Changes

This PR does the following:
* Refactors the `ParameterEditor` component to default to a new `DefaultPeriodSetter` component that is very similar to our current default, but allows the selection of a start year
* Adds explicity "yearly" and "date" input methods to replace our existing popover
* Adds a new "multi-year" method that allows for annual input of up to ten years at once
* Adds a new popover via a gear icon to allow a user to select one of these input methods

## Screenshots

https://github.com/user-attachments/assets/7e4d3a94-182e-456b-84ed-240ada7c831b

## Tests

This PR does not add tests. I've found Jest to often inhibit the building of its own tests, and would recommend moving to using it only for unit testing individual functions, then using Playwright for end-to-end testing that would include the `ParameterEditor` component.

### DIFF
--- a/src/__tests__/algorithms/Bisection.test.js
+++ b/src/__tests__/algorithms/Bisection.test.js
@@ -1,110 +1,31 @@
 import { bisect } from "algorithms/Bisection";
 describe("bisection", function () {
-  test("small cases", function () {
-    expect(
-      bisect([], 1, function (a) {
-        return a - 1;
-      }),
-    ).toBe(0);
-    expect(
-      bisect([1], 1, function (a) {
-        return a - 1;
-      }),
-    ).toBe(0);
-    expect(
-      bisect([1], 2, function (a) {
-        return a - 2;
-      }),
-    ).toBe(1);
-    expect(
-      bisect(
-        [1],
-        1,
-        function (a) {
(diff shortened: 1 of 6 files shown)


anth-volk:
@nikhilwoodruff Do you mean that the input values are slightly lower than the years? Or do you mean that the years themselves don't align with "Current value"? 

If the latter, I had purposely done that because the inputs inside the yearly and date-based input are also 12px to the right, due to the padding inside of their respective `DatePicker` and `YearPicker` components. I could record how this looks when switching between items, as I think it makes switching look a bit cleaner.

nikhilwoodruff:
Ok nvm then- good to merge! Makes sense.

PR #2140, opened 2024-10-27
fixed scroll to top developers page

### LINKED ISSUE #2132: After clicking link to Developer Tools from footer, user should automatically be scrolled to top of page
I think we add one line of code elsewhere to ensure this happens - should be as simple as copying and pasting that

bharateshwq:
## Description
Fixes #2132
Developer page scrolls to top when mounted

## Changes

Restructured the page along with a side effect to scroll to 0,0


anth-volk (changes requested):
Hi @bharateshwq, thanks for your contributions here. Just two quick things:
1. Curious why you chose to separate out the main content into its own arrow function. Is that to better align with something? If not, I think it may be preferable to leave it as it is.
2. Could you lint and push?

anth-volk:
Closing due to inactivity.

PR #2166, opened 2024-10-31
Benefits uprating 2025/26 post

vahid-ahmadi:
I write a post about the 1.7% benefits uprating for 2025/26 in the UK.


vahid-ahmadi:
Fixes #2167

PR #2213, opened 2024-11-19
Make plus button blue

### LINKED ISSUE #2199: Change color of policy selector button to `BLUE_PRIMARY`
Missed this previously when changing button colors. It should remain a secondary button, but become filled in on hover.

Screen Shot 2024-11-08 at 12 38 29 AM

anth-volk:
## Description

Fixes #2199.

## Changes

Makes the plus button in the policy selector component blue instead of teal.

### DIFF
--- a/src/controls/Button.jsx
+++ b/src/controls/Button.jsx
@@ -39,6 +39,18 @@ export const buttonStyles = {
       color: colors.WHITE,
     },
   },
+  secondaryBlue: {
+    standard: {
+      backgroundColor: "transparent",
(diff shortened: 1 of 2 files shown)


MaxGhenis:
out of curiosity could you show how it looks here and with both the check and plus as teal?

anth-volk:
And then as blue:
[AwesomeScreenshot-11_19_2024,11_07_36PM.webm](https://github.com/user-attachments/assets/af0cf5c1-7248-4e6d-b7dc-ee6c97d81af7)

I would also advocate for the rename button to be changed to blue

Issue #2263, opened 2024-12-19
App should forward authentication token to the API on all requests


mikesmit:
Currently we only use the authentication token from auth0 in the app itself. We should also be passing it to the back end API.

The scope of this task is to investigate the correct way to do that and then add the correct token to the request.

mikesmit:
Some context for development

Auth0 is configured in [Auth0ProviderWithNavigate.jsx](https://github.com/PolicyEngine/policyengine-app/blob/ff9a7a142b10faf037566a9d31dae11312335b7d/src/auth/Auth0ProviderWithNavigate.jsx#L5) by creating the [Auth0Provider](https://auth0.github.io/auth0-react/functions/Auth0Provider.html)

```
return (
    <Auth0Provider
      domain={domain}
      clientId={clientId}
      authorizationParams={{
        redirect_uri: redirectUri,
      }}
      onRedirectCallback={onRedirectCallback}
      useRefreshTokens={true}
      cacheLocation="localstorage"
    >
      {children}
    </Auth0Provider>
```

Required authentication  is done via the [AuthenticationGuard.jsx](https://github.com/PolicyEngine/policyengine-app/blob/master/src/auth/AuthenticationGuard.jsx#L6-L12) which in turn uses [auth0-react's withAuthenticationRequired](https://auth0.github.io/auth0-react/functions/withAuthenticationRequired.html)

```
export const AuthenticationGuard = ({ component, ...props }) => {
  const Component = withAuthenticationRequired(component, {
    onRedirecting: () => <AuthCallback />,
  });

  return <Component {...props} />;
};
```

Optional authentication is done with the [useAuth0 hook](https://auth0.github.io/auth0-react/functions/useAuth0.html)

one example of this is [in PolicyEngine.jsx](https://github.com/PolicyEngine/policyengine-app/blob/ff9a7a142b10faf037566a9d31dae11312335b7d/src/PolicyEngine.jsx#L194)

mikesmit:
We should replace [the places we use fetch](https://github.com/search?q=repo%3APolicyEngine%2Fpolicyengine-app+fetch%28&type=code) ideally with a single mechanism that gets the auth0 token (if it exists) and adds it to a header "authorization" in the request.

mikesmit:
[the auth0 documentation](https://auth0.com/docs/quickstart/spa/react/02-calling-an-api) just uses fetch directly and is not integrated with auth0-react.

[...]

PR #2264, opened 2024-12-19
Add bearer token just to user-profile

mikesmit:
partially adresses #2263

## Description

In order for us to authorize calls to the API, our app needs to send authentication information to the API. This change incrementally adds an authentication token to some of the requests to the API. Further changes will add that token to all requests in the app and then changes to the API will use the token to authorize requests where appropriate.

## Changes

1. Adds a new hook useAuthentcatedFetch which provides a version of fetch that includes the Authentication header when the user is logged in
2. Adds a new useAuthenticatedApiCall which provides a wrapped apiCall using that new authentication header
3. Uses that new authenticatedApiCall for the PUT/POST calls for user-profile

Next step will be to integrate with other calls in the app.

I also did not attempt to add component integration tests where they do not already exist.
## Screenshots

Visually nothing has changed

## Tests

I added unit tests, but I also ran the page locally (manually changing the auth redirect to go to localhost:3000) to verify that the network requests now include an authorization header with a bearer token in them.

### DIFF
--- a/src/PolicyEngine.jsx
+++ b/src/PolicyEngine.jsx
@@ -16,10 +16,10 @@ import { COUNTRY_BASELINE_POLICIES, COUNTRY_CODES } from "./data/countries";
 
 import { useEffect, useState, lazy, Suspense } from "react";
 import {
-  apiCall,
   copySearchParams,
   countryApiCall,
   updateMetadata,
+  useAuthenticatedApiCall,
 } from "./api/call";
 import LoadingCentered from "./layout/LoadingCentered";
 import ErrorPage from "./layout/ErrorPage";
@@ -103,6 +103,8 @@ export default function PolicyEngine() {
 
   const [hasShownHouseholdPopup, setHasShownHouseholdPopup] = useState(false);
   const [userProfile, setUserProfile] = useState({});
+
+  const { authenticatedApiCall } = useAuthenticatedApiCall();
 
   // Update the metadata state when something happens to
   // the countryId (e.g. the user changes the country).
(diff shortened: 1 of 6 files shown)


src/hooks/useAuthenticatedFetch.js
```diff
@@ -0,0 +1,38 @@
+import { useAuth0 } from "@auth0/auth0-react";
+import { useCallback } from "react";
+
+/**
+ * Get an 'authenticatedFetch' function which, if the user is logged in,
+ * will automatically attach an access token to any API request.
+ * @returns {{authenticatedFetch:(path:string, opts:Record)=>Promise}}
+ */
+export function useAuthenticatedFetch() {
```
anth-volk, line 9: __accolade:__ I appreciate the use here of an authenticated fetch function more generally, that is then utilized by the authenticated API call function

src/PolicyEngine.jsx
```diff
@@ -239,7 +245,7 @@ export default function PolicyEngine() {
     if (countryId && isAuthenticated && user?.sub) {
       fetchUserProfile().then((userProfile) => setUserProfile(userProfile));
     }
-  }, [countryId, user?.sub, isAuthenticated]);
+  }, [countryId, user?.sub, isAuthenticated, authenticatedApiCall]);
```
anth-volk, line 248: __question, non-blocking:__ Theoretically, at a future step, would we expect to extend `authenticatedApiCall` to all calls, then be required to add the function to all effect hook dep arrays?

Or perhaps another way to frame this: we clearly have too many effect hook calls

anth-volk (approved):
@mikesmit This all looks great, thank you for this contribution. I realize that implementing token insertion within `user_profile` calls is a first-round effort toward implementing more broadly across all endpoints.

I'll leave this unmerged so that you can merge alongside any API changes you might like to make first, if desired.

src/PolicyEngine.jsx
mikesmit: Yes, my plan is to switch everything over to auto-authenticated API calls and that will result in everything making an API call using a hook.

I haven't taken a deep dive into the react and use of hooks yet, but I agree there is too much in general going on in our react components.

PR #2291, opened 2025-01-03
Use Authorization header instead of Authentication

mikesmit:
partially addresses #2263
    
updated useAuthenticatedFetch to correctly set the "Authorization" header to the bearer token instead of
"Authentication"


MaxGhenis:
Think the title should be flipped?

mikesmit:
> Think the title should be flipped?

Demonstrating how this happened in the first place.... Fixing title.

anth-volk:
Just found where auth0 specifies this in the docs [here](https://auth0.com/docs/quickstart/backend/python/interactive). Thanks, will approve and merge!

anth-volk (approved):

PR #2316, opened 2025-01-23
Point reproduce in Python segments to HuggingFace URLs

### LINKED ISSUE #2315: Point reproduce in Python segments to HuggingFace URLs
These will more accurately/strongly enforce data parity with the API.

nikhilwoodruff:
Fixes #2315

## Description

Point dataset strings to URLs instead

## Changes

As above.

## Screenshots

image
## Tests

No tests added.

### DIFF
--- a/src/data/reformDefinitionCode.js
+++ b/src/data/reformDefinitionCode.js
@@ -189,9 +189,9 @@ export function getImplementationCode(
   if (hasDatasetSpecified) {
     datasetText = DEFAULT_DATASETS[dataset];
   } else if (isState) {
-    datasetText = "pooled_3_year_cps_2023";
+    datasetText = "hf://policyengine/policyengine-us-data/pooled_3_year_cps_2023.h5";
   } else if (region === "enhanced_us") {
-    datasetText = "enhanced_cps_2024";
+    datasetText = "hf://policyengine/policyengine-us-data/enhanced_cps_2024.h5";
   }
 
   const datasetSpecifier = datasetText ? `dataset="${datasetText}"` : "";
--- a/src/data/countries.js
+++ b/src/data/countries.js
@@ -58,7 +58,7 @@ export const STATUS_TEXT_COLORS = {
 // this only applies to the "enhanced_cps"
 // dataset selection
 export const DEFAULT_DATASETS = {
-  enhanced_cps: "enhanced_cps_2024",
+  enhanced_cps: "hf://policyengine/policyengine-us-data/enhanced_cps_2024.h5",
 };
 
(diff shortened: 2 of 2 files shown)


src/data/reformDefinitionCode.js
```diff
@@ -189,9 +189,9 @@ export function getImplementationCode(
   if (hasDatasetSpecified) {
     datasetText = DEFAULT_DATASETS[dataset];
   } else if (isState) {
-    datasetText = "pooled_3_year_cps_2023";
+    datasetText = "hf://policyengine/policyengine-us-data/pooled_3_year_cps_2023.h5";
```
anth-volk, line 192: __question, blocking:__ Would it be more sustainable to modify this setting in `us-data` and keep the app the same?

It seems to me we always want the API and country packages both to prefer HuggingFace datasets over GitHub. It also seems like it'd be relatively simple to modify `us-data` to either recognize this preference, or just only download from once source or the other, instead of our current method where we still send data updates to both GitHub and HuggingFace. 

Could we just modify `us-data` to do this? Seems unwieldy to ask users to remember this entire link.

anth-volk (commented):
Hey @nikhilwoodruff, thanks for this quick turnaround. My main thought is in my comment below, but TL;DR, would it be more logical (and equally as fast) to solve this in `us-data`? Looking forward to your thoughts.

src/data/reformDefinitionCode.js
nikhilwoodruff: As per the policyengine.py design doc, I think we should move away from us-data being a package dependency so I'd vote against this.

In the medium term we'll move to policyengine.py, which abstracts the huggingface URL away behind constants, and update the reproduce in python segment accordingly. Right now the reproduce in python segment is broken without this, so I'd suggest still keeping this solution in the -app repo here.

anth-volk: Okay, I have a better understanding now. This seems sensible as a temporary change. @PavelMakarchuk and @DTrim99 Just to flag for you - we'll temporarily be listing this longer URL, but will revert to the typical method of choosing datasets after integrating `policyengine.py` more broadly.

__question:__ Would we need to also do this for the standard CPS? Is that not defaulting to pulling outdated data from GitHub?

anth-volk (approved):
Based on conversation with @nikhilwoodruff, I think this is good to go, as we will re-enable simpler dataset loading with the inclusion of `policyengine.py` within the API. @nikhilwoodruff if you could lint this, it should be good to merge!

anth-volk:
@nikhilwoodruff I've taken the liberty of linting in order to get this merged.

Issue #2335, opened 2025-02-02
Remove duplicate color names


MaxGhenis:
e.g. BLUE_PRIMARY and BLUE are both #2C6496

https://github.com/PolicyEngine/policyengine-app/blob/master/src/style/colors.js

Rohithvema79:
Hi,
BLUE_98 (#F7FAFD) and TEAL_LIGHT (#F7FDFC) are very close to each other and there are no exact duplicates in colors.js

SEEDART007:
@MaxGhenis hey kindly close this issue its fixed now

PR #2350, opened 2025-02-08
fix: Remove duplicate color names #2335

Devansh633:
Removed PRIMARY_BLUE from style/color.js 
Replaced PRIMARY_BLUE to BLUE in all files


MaxGhenis:
could you also unify BLUE95 and BLUE_LIGHT?

Devansh633:
Sure, I'll do that

anth-volk (approved):
Thanks for this @Devansh633! Good to merge.

PR #2371, opened 2025-02-24
Fixed displayed values ordering

### LINKED ISSUE #2342: Displayed values seem to have no order
For some reason, in various places, values read from the API appear in no particular order, like the below. I suspect the cause of this is a change in the API, but opening here for the time being, as its ramifications are visible in the front end.

Image

harnish7576:
## Description
Fixes #2342 
- Sorting naively created sorted list which had entirely new objects. 
- **SortTreeInPlace** happens before rendering, modifying the original objects.
- This way it only affects the order, not object structure. 
- **StackedMenu** now receives the same object references which preserves interactivity.

## Changes
- Added a **SortTreeInPlace** Function that sorts the same objects in memory instead of creating new ones and Ensures children are sorted recursively.
- Sorted based on Labels instead of Name

## Screenshots
Screenshot 2025-02-24 at 2 56 02 PM

Screenshot 2025-02-24 at 2 56 45 PM


anth-volk:
Thanks for your work on this @harnish7576! A couple things:

* Please add "Fixes #NUMBER" as the first item in the PR description
* Please rename the issue as something readable (e.g., "Fixed displayed values ordering")

harnish7576:
@anth-volk  I've updated the PR as requested.

src/pages/PolicyPage.jsx
```diff
@@ -67,6 +67,25 @@ function PolicyLeftSidebar(props) {
...
+      if (Array.isArray(item.children)) {
+        sortTreeInPlace(item.children);
+      }
+    });
+
+    return tree;
+  };
+
+  sortTreeInPlace(metadata.parameterTree.children);
+  if (POLICY_OUTPUT_TREE.length > 0) {
```
anth-volk, line 85: __issue, blocking:__ We're only looking to sort the metadata parameter tree

We actually have a custom-defined order on the outputs, so please only sort `metadata.parameterTree.children`

src/pages/PolicyPage.jsx
```diff
@@ -67,6 +67,25 @@ function PolicyLeftSidebar(props) {
     window.location.search.includes("focus=policyOutput") ||
     window.location.search.includes("focus=householdOutput");
 
+  const sortTreeInPlace = (tree) => {
+    if (!Array.isArray(tree)) return [];
+
+    tree.sort((a, b) => a.label.localeCompare(b.label));
```
anth-volk, line 73: __accolade:__ Like the use of localCompare and the clean sort on this

anth-volk (changes requested):
Hey @harnish7576, thanks for this! I've requested one minor change, then should be good to merge.

anth-volk (approved):
Thank you for this @harnish7576! Glad to get this merged.

Issue #2099, opened 2024-10-15
Properly handle 404's from `get_policy_search` endpoint


anth-volk:
In conjunction with RESTful expectations, https://github.com/PolicyEngine/policyengine-api/pull/1854 returns a 404 when no records are found from the `get_policy_search` endpoint. However, that's not necessarily an indication of failure, just that no records have been found. We'll need to update the policy search box to allow 404's and just to output no policies, instead of erroring, which I believe is only console logged at the moment in prod, but throws a runtime error while testing.

anth-volk:
Closed by #2388.

PR #2388, opened 2025-03-03
Fixing policy search to handle 404/Invalid Search API responses

harnish7576:
## Description
Issue #2099 
- 404 responses indicate no matching policies were found (not a failure case)
- The component was treating this as an error condition

## Changes
- Updated the policy search component to specifically handle 404 responses
- When a 404 is encountered, the component now sets an empty policies array instead of throwing an error

## Screenshots
Screenshot 2025-03-03 at 6 08 45 PM


anth-volk:
Could you change `Issue #2099` to `Fixes #2099`? Otherwise the auto-close tool will not work.

PR #2391, opened 2025-03-05
AI landing page

### LINKED ISSUE #2390: AI landing page

MaxGhenis:
Fixes #2390


anth-volk:
Will review today, but beforehand, this fails lint due to unused variables.

src/layout/Header.jsx
```diff
@@ -402,54 +551,216 @@ function PageLinks() {
...
+}
+
+function LeftNavigationMenu(props) {
+  // The menu that slides in from the left when the hamburger is clicked
+  const { isOpen } = props;
+  const countryId = useCountryId();
+  const [expandedDropdown, setExpandedDropdown] = useState(null);
+
+  const renderMobileLink = (link, i) => {
+    if (link.isDropdown) {
```
anth-volk, line 566: __issue, blocking:__ This, too, contains logic that is hard to follow

Here, too, I'd recommend breaking into sub-components and making `LeftNavigationMenu` a vehicle for mapping links into visual components

src/layout/Header.jsx
```diff
@@ -402,54 +551,216 @@ function PageLinks() {
...
+      );
+    }
+  };
 
-function LeftNavigationMenu(props) {
-  // The menu that slides in from the left when the hamburger is clicked
-  const { isOpen } = props;
-  const countryId = useCountryId();
+  // Reset expanded dropdown when menu closes
+  useEffect(() => {
```
anth-volk, line 759: __suggestion:__ Use conditional rendering instead of an effect hook

This should also be as simple as returning `null` if `isOpen` is falsy, and more generally, I feel we should use effect hooks as little as possible, as they provide a performance hit due to forced re-paint and create side effects that are sometimes necessary, but can be difficult to follow at times

src/pages/AIGeneralContent.jsx
```diff
@@ -0,0 +1,288 @@
...
+
+  // Configure content specific to the AI page
+  const aiContent = {
+    title: "AI & Machine Learning",
+    subtitle:
+      "PolicyEngine uses artificial intelligence and machine learning to make policy analysis more accurate and accessible.",
+    heroTitle: "Making Policy Accessible With AI",
+    heroSubtitle: `PolicyEngine combines tax-benefit microsimulation with cutting-edge AI to ${democratize} policy understanding`,
+    heroImage: householdImage,
+    heroButtonText: "Try AI Explanations",
```
anth-volk, line 30: __question:__ Is this section descriptive enough?

At the moment, "Try AI Explanations" links to just a standard household input, and we provide no walk-through this high up on the page as to what our AI features do. To me as a potential user, I wouldn't understand what to do next.

src/pages/AIGeneralContent.jsx
```diff
@@ -0,0 +1,288 @@
...
+              <p>
+                For policy analysis, we leverage GPT-4 to weave narratives from
+                our computational results, explaining reforms in terms anyone
+                can understand - from simplified &quot;ELI5&quot; explanations
+                to detailed technical analyses for policy experts.
+              </p>
+            </div>
+            <div className="how-it-works-image">
+              <img
+                src={require("../images/posts/ai-analysis.png")}
```
anth-volk, line 140: __nit:__ This image shows the OpenAI logo, but we use Claude for this feature

src/pages/AIGeneralContent.jsx
```diff
@@ -0,0 +1,288 @@
...
+      >
+        {/* Additional AI-specific content */}
+        <section className="section">
+          <div className="section-content">
+            <h2>Try PolicyEngine AI</h2>
+            <div
+              className="demo-button-container"
+              style={{ textAlign: "center", margin: "2rem 0" }}
+            >
+              <h3>Try the AI explanation feature</h3>
```
anth-volk, line 167: __suggestion:__ I feel the styling on this section is lacking

We basically place an h3 header under an h2 header, then create a button that mirrors none we use anywhere else

src/pages/AIGeneralContent.jsx
```diff
@@ -0,0 +1,288 @@
...
+              <HouseholdAIModal isDemo={true} />
```
anth-volk, line 168: __question, blocking:__ Is this feature useful here?

This segment of the page only shows a sample AI output, but does not walk a user through a sample household input that would generate it or link this to the household input experience. If I saw this and shows the modal, I wouldn't understand the purpose.

src/pages/AIGeneralContent.jsx
```diff
@@ -0,0 +1,288 @@
...
+                frameBorder="0"
+                allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
+                allowFullScreen
+              ></iframe>
+            </div>
+          </div>
+        </section>
+      </GeneralContent>
+
+      <style>{`
```
anth-volk, line 191: __nit:__ We don't use this form of styling anywhere else in the app

Can we keep it, especially in light of an upcoming comprehensive redesign? Sure. This styling's also arguably a better system than the inline we're using elsewhere. Even so, it doesn't match anywhere else in the code base.

src/pages/APIDocumentationPage.jsx
```diff
@@ -97,35 +91,7 @@ export const exampleInputs = {
...
-auth_response = requests.post(\"https://policyengine.uk.auth0.com/oauth/token\", headers=headers, json=payload)
-
-result = auth_response.json()
-print(result[\"access_token\"])`;
-
-const tokenOutputCode = `{
-  "access_token": "YOUR_ACCESS_TOKEN",
-  "token_type": "Bearer"
-}`;
+// Note: These constants are kept for reference but not currently used
```
anth-volk, line 94: __suggestion:__ Move these constants somewhere else, as about 70% of this file is just unused constants

src/pages/APIGeneralContent.jsx
```diff
@@ -0,0 +1,324 @@
...
+
+result = auth_response.json()
+print(result["access_token"])`;
+
+  const tokenOutputCode = `{
+  "access_token": "YOUR_ACCESS_TOKEN",
+  "token_type": "Bearer"
+}`;
+
+  const exampleHouseholdInput = {
```
anth-volk, line 40: __question:__ Is this not the same example input as that in `APIDocumentationPage`?

src/pages/APIGeneralContent.jsx
```diff
@@ -0,0 +1,324 @@
...
+  const apiContent = {
+    title: "API Documentation",
+    subtitle:
+      "Access PolicyEngine's powerful tax-benefit microsimulation engine programmatically",
+    heroTitle: "PolicyEngine REST API",
+    heroSubtitle:
+      "Build tax and benefit policy analysis into your applications with our comprehensive API",
+    heroImage: require("../images/home/api_screenshot.png"),
+    heroButtonText: "Try API Playground",
+    heroButtonLink: `#api-playground`,
```
anth-volk, line 88: __issue:__ This button doesn't work for me

src/pages/APIGeneralContent.jsx
```diff
@@ -0,0 +1,324 @@
...
+        id: "variables",
+        content: (
+          <div className="api-section">
+            <p>
+              Access information about all available variables and parameters in
+              the PolicyEngine API. This endpoint returns metadata about all the
+              inputs and outputs that can be used in calculations.
+            </p>
+            <h5>
+              <code>GET /{countryId}/metadata</code>
```
anth-volk, line 206: __suggestion:__ This `code` segment feels abrupt here; perhaps some textual changes on the lines above could improve it, or perhaps moving the `code` segment up in anticipation of adding new API endpoints

src/pages/BenefitAccessPage.jsx
```diff
@@ -0,0 +1,392 @@
...
+  const countryId = useCountryId();
+  // Country identification logic
+
+  useEffect(() => {
+    document.title = "Benefit Access | PolicyEngine";
+    window.scrollTo(0, 0);
+  }, [location]);
+
+  const generalPages = [
+    { id: "ai", label: "AI & ML", path: `/${countryId}/ai` },
```
anth-volk, line 26: __question:__ Is this bar not built into the page already in a component sitting above this?

src/pages/BenefitAccessPage.jsx
```diff
@@ -0,0 +1,392 @@
...
+    {
+      id: "opensource",
+      label: "Open Source",
+      path: `/${countryId}/opensource`,
+    },
+  ];
+
+  // Partner organizations using PolicyEngine API for benefit access tools
+  const partnerOrganizations = [
+    {
```
anth-volk, line 48: __suggestion:__ Refactor this data to sit in an external JS or JSON file, then import to separate data from styling & logic

src/pages/BenefitAccessPage.jsx
```diff
@@ -0,0 +1,392 @@
...
+        "https://static1.squarespace.com/static/5f6287b7ba52c722644c43c2/t/5f87a74f3990a2436a9f1b56/1740173730712/",
+      regions: ["Los Angeles County"],
+    },
+    {
+      name: "Navvy",
+      description:
+        "A benefits screener developed by Impactica for the Student Basic Needs Coalition. Navvy helps students identify and access critical support services and benefits specifically designed for higher education students.",
+      website: "https://www.studentbasicneeds.com/navvy",
+      imageUrl:
+        "https://static.wixstatic.com/media/bc112d_b0a5a365b2ec4fa587724e776fa99934~mv2.png/v1/crop/x_0,y_10,w_1436,h_1322/fill/w_832,h_766,al_c,q_90,usm_0.66_1.00_0.01,enc_avif,quality_auto/unnamed-2_edited.png",
```
anth-volk, line 76: __question:__ Can we host these images locally to avoid future link breakage?

src/pages/BenefitAccessPage.jsx
```diff
@@ -0,0 +1,392 @@
...
+              color: style.colors.BLACK,
+            }}
+          >
+            {page.label}
+          </Link>
+        ))}
+      </div>
+
+      <Section>
+        <div className="benefits-intro">
```
anth-volk, line 128: __issue:__ This page section looks fundamentally different from others in this navbar

More fundamentally: do we need a page header, then a mini navbar, then a page section that functions as another page header?

src/pages/BenefitAccessPage.jsx
```diff
@@ -0,0 +1,392 @@
...
+                  <img src={org.imageUrl} alt={`${org.name} screenshot`} />
+                </div>
+              )}
+              <a
+                href={org.website}
+                target="_blank"
+                rel="noopener noreferrer"
+                className="partner-website"
+              >
+                Visit Website →
```
anth-volk, line 179: __issue:__ Nikhil will flip if he sees capitalization like this

src/pages/BenefitAccessPage.jsx
```diff
@@ -0,0 +1,392 @@
...
+        }
+        
+        .partner-website:hover {
+          color: ${style.colors.DARK_BLUE_HOVER};
+        }
+        
+        .cta-section {
+          max-width: 800px;
+          margin: 0 auto;
+          text-align: center;
```
anth-volk, line 334: __issue:__ We don't really center text in most places, and the way this is centered (with a very small `max-width` makes the text look strange on standard desktop sizes

src/pages/BenefitAccessPage.jsx
```diff
@@ -0,0 +1,392 @@
...
+        .cta-buttons {
+          display: flex;
+          gap: 1rem;
+          justify-content: center;
+          margin-top: 2rem;
+        }
+        
+        .cta-button {
+          padding: 0.8rem 1.5rem;
+          border-radius: 4px;
```
anth-volk, line 356: __issue:__ We don't use `border-radius` anywhere else

I'd recommend using in future, but here, this looks jarringly different

src/pages/EducationPage.jsx
```diff
@@ -0,0 +1,142 @@
...
+            If you&apos;re using PolicyEngine in an educational setting and
+            would like to contribute your story or resources, please contact us
+            at{" "}
+            <a href="mailto:hello@policyengine.org">hello@policyengine.org</a>.
+          </p>
+        </div>
+      </Section>
+
+      <style>{`
+        .placeholder-content {
```
anth-volk, line 115: __issue:__ This strongly differs from our styling elsewhere; recommend overhauling

src/pages/GeneralContent.jsx
```diff
@@ -0,0 +1,203 @@
...
+ * @param {Object} props.content - Content configuration specific to the page type
+ * @param {React.ReactNode} props.children - Additional content specific to the page type
+ */
+const GeneralContent = ({
+  countryId,
+  isUK = false,
+  pageType,
+  content = {},
+  children,
+}) => {
```
anth-volk, line 26: __issue:__ I get where this component is going, but it seems like a strange intermediate item when we already have a custom `Section` component

Also, having default filled values could cause us to miss cases where we need to override, but forget to do so

src/pages/GeneralContent.jsx
```diff
@@ -0,0 +1,203 @@
...
+    heroButtonText: "Learn More",
+    heroButtonLink: `/${countryId}`,
+    sections: [], // Will be filled by specific page implementations
+  };
+
+  // Merge provided content with defaults
+  const pageContent = { ...defaultContent, ...content };
+
+  // Define general navigation
+  const generalPages = [
```
anth-volk, line 47: __question:__ Is this also meant to be overridden? Why not just take this as a prop and default to empty when not provided?

src/pages/GeneralContent.jsx
```diff
@@ -0,0 +1,203 @@
...
+  ];
+
+  return (
+    <div>
+      <PageHeader title={pageContent.title} backgroundColor="#F7FAFD">
+        <p style={{ margin: 0 }}>{pageContent.subtitle}</p>
+      </PageHeader>
+
+      {/* General navigation tabs */}
+      <div
```
anth-volk, line 75: __thought:__ I'm not a fan of having a navbar, underneath which is a header, underneath which is a separate navbar, but unsure on the solution here; would lean toward just not having, since the main navbar contains a dropdown menu (with slightly different items) that contains all items on the second nav menu

src/pages/GeneralContent.jsx
```diff
@@ -0,0 +1,203 @@
...
+          <div className="hero-content">
+            <h2>{pageContent.heroTitle}</h2>
+            <p className="hero-subtitle">{pageContent.heroSubtitle}</p>
+            <div className="hero-buttons">
+              <LinkButton
+                link={pageContent.heroButtonLink}
+                text={pageContent.heroButtonText}
+                type="primary"
+              >
+                <FontIcon name="calculator" />
```
anth-volk, line 116: __question:__ What does this do?

src/pages/MicrosimContent.jsx
```diff
@@ -0,0 +1,307 @@
...
+    heroSubtitle: `PolicyEngine uses microsimulation models to ${analyze} the impact of policy reforms on real households`,
+    heroImage: require("../images/posts/how-machine-learning-tools-make-policyengine-more-accurate.png"),
+    heroButtonText: "Try Policy Analysis",
+    heroButtonLink: `/${countryId}/policy`,
+    sections: [
+      {
+        title: "What is Microsimulation?",
+        id: "microsim-intro",
+        content: (
+          <div className="microsim-intro">
```
anth-volk, line 30: __issue:__ Here, again, the centering of this component, followed by non-centering of the next section, make this look very strange

src/pages/AIGeneralContent.jsx
MaxGhenis: could definitely be better but would like to get something out that we can build on. we could replace it with a link to the blog post though

src/pages/AIGeneralContent.jsx
MaxGhenis: sure we could remove it

anth-volk:
@MaxGhenis I've made some modifications to the component, and I'd say it's now in a mergeable state in the context of a short-turnaround, short-term set of components. However, I would advocate for a pretty significant refactor on this in the future as part of a redesign. I'm going to note down some highlights in a separate comment for whoever might take this on in the future.

anth-volk:
Some considerations when refactoring:
1. The components utilize a pretty atypical architecture. Ideally, we'd expect a sort of "page" component that defines some HTML metadata, then implements a series of modular "section" components that handle their own styling and take input via props that will then be displayed in an orderly fashion. 

These components, instead, have a "page" that handles HTML metadata only, then (sometimes) return inside of that a `GeneralComponent` that is not modular enough and far too strongly tied to a particular implementation. This, in turn implements the "section" components described before, but in a way that mixes logic and display. These components also utilize a wonky inline styling method that is not currently recommended.

2. The components seem to rely upon a styling system that does not 100% match our own. The worst of this is probably the secondary navbar that these components add, as well as the secondary header that all `GeneralComponent` items must have, complete with a required CTA and link button, which doesn't fit the needs of some of these pages.

3. These components tend to avoid modularization, instead favoring long implementations that mix logic, styling, and data in ways that make them hard to maintain and debug. This is particularly true of the secondary navbar. If we decide to keep it (which I'd strongly recommend against), the secondary navbar will need to be further broken up.

anth-volk (approved):
Approving with the following provisos:

1. We want something with a very quick turnaround
2. We will be refactoring these as part of ongoing redesign efforts

PR #2430, opened 2025-03-21
Use UTC year for date inputs

### LINKED ISSUE #2417: Change default parameter change as-of year from 2024 to 2025

anth-volk:
## Description

Fixes #2417.

## Changes

The `DefaultParameterEditor` component uses JavaScript's `Date` constructor to take an ISO date string `startDate` prop and convert it to a JS date-type. However, the standard constructor is time zone-sensitive and assumes date strings without a time zone are in UTC. Thus, when we passed `2025-01-01` to `DefaultParameterEditor`, this was converted for users in the Americas to `December 31, 2024, 7pm`. We then parsed the year off this value using the `getFullYear` method.

This PR replaces the use of `getFullYear` with `getUTCFullYear`, which corrects this problem by treating all `Date` instances as UTC.

## Screenshots
Screen Shot 2025-03-21 at 5 46 57 PM

## Tests

None have been added.

### DIFF
--- a/src/data/constants.js
+++ b/src/data/constants.js
@@ -4,7 +4,7 @@
 // allow for application into perpetuity, if that is desired behavior
 export const defaultForeverYear = "2100";
 
-export const defaultYear = new Date().getFullYear();
+export const defaultYear = new Date().getUTCFullYear();
 export const defaultStartDate = defaultYear.toString().concat("-01-01");
 export const defaultEndDate = defaultForeverYear.toString().concat("-12-31");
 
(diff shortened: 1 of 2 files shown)


anth-volk:
Forgot to stay on top of shepherding this through review. Requested @mikesmit's review, as well, since this is a minor change.

mikesmit (commented):
Verified these are the only two cases of getFullYear in the app, but I also saw this: https://github.com/PolicyEngine/policyengine-app/blob/d46cbcabaf2f6999807e9b7aa02b3dd9d8747f7e/src/lang/format.js#L85-L90

Do we need to change that too?

anth-volk:
I've flagged your comment for later review, Mike, as part of #2447. Thanks for review!

Issue #2471, opened 2025-04-03
Add function to aggregate outputs for display in society-wide outputs page


anth-volk:
This function should do the following:

Given an instance of the society-wide structure we currently use, aggregate the values from the structures to create one aggregated instance of the same schema; this may be challenging, as for some values, we'll need the aggregate; for some, we'll need mean; and for others, perhaps something else?

I will want to confirm with Pavel or Nikhil how they calculate multi-year impacts just to be 100.0% certain this function is doing the math properly.

- [x] Create function to aggregate a series of SocietyWideImpact items into one aggregated total output
- [x] Write tests for the above function
- [x] Modify function to remove unneeded outputs - only want to aggregate top-line items - we only want the data needed for the overview page
- [x] Create schema representing aggregated impacts
- [x] Re-open PR off of a clone instead of a fork

PavelMakarchuk:
For the common web app outputs:
Budgetary impact: sum 
Distributional impacts / Winners and Losers / Poverty / Inequality : mean 

Unsure whether we want to and how we would handle the cliff impacts and LSRs (will take over an hour in the current state)

PR #2498, opened 2025-04-15
Create schematic representations of US and UK society-wide outputs

anth-volk:
## Description

Fixes #2483. Re-opened version of https://github.com/PolicyEngine/policyengine-app/pull/2491 and of https://github.com/PolicyEngine/policyengine-app/pull/2497. Requires merging of https://github.com/PolicyEngine/policyengine-app/pull/2476.

Note for reviewer: I am shakier on the UK schema, and would love more thorough review there. Do all runs expose constituency-level data? Is the UK program module correct here?

## Changes

This PR creates a schematic representation of society-wide outputs for US and UK app runs. It does not do so for any other country case, as they do not provide microdata, and thus do not enable society-wide runs.

## Screenshots

N/A

## Tests

None have been added, as these are aimed at representing schemas.

PR #2499, opened 2025-04-15
Add functionality to aggregate multi-year outputs

anth-volk:
## Description

Fixes https://github.com/PolicyEngine/policyengine-app/issues/2471. Requires merging of #2498. 

## Changes

Adds a series of functions to aggregate a society-wide impact result for the US or UK, as well as a series of tests.

## Screenshots

N/A

## Tests

A bevy have been added to confirm the accurate functioning of the functions and their sub-components.


anth-volk:
Note: Automated tests do not run for chained PRs, since these are not opened against `master`. Please ignore the "success" noted for the PRs here, this does NOT mean that tests have passed.

src/api/societyWideAggregation/aggregateModules.js
```diff
@@ -0,0 +1,47 @@
...
+      budgets.map((b) => b?.tax_revenue_impact),
+    ),
+  };
+}
+
+export function aggregateIntraDecileModule(intraDecileData) {
+  return {
+    all: aggregateWinnersLosersBreakdownSimple(
+      intraDecileData.map((d) => d?.all),
+      "mean",
```
nikhilwoodruff, line 31: I don't see a big argument for using `mean` over `last` (cc @maxghenis) but not a big deal either way as long as we're clear which we're using in the UI.

nikhilwoodruff (approved):

src/api/societyWideAggregation/aggregateModules.js
nikhilwoodruff: Reforms which phase in over time, for example, would have a bigger impact in the winners/losers chart

MaxGhenis: Taking a step back - I thought we agreed we would only show budgetary impacts over 10y

anth-volk: This PR was intended to just be a series of functions to aggregate relevant API output when I thought we wanted far more than we did. https://github.com/PolicyEngine/policyengine-app/pull/2502 was then meant to create all the display features. All aggregators other than budgetary are unused in #2502. I could remove them from the PR if desired.

MaxGhenis: Yes please remove them. All distributional analyses I've seen are as of a single year. Sorry for any miscommunication.

MaxGhenis (approved):

src/api/societyWideAggregation/aggregateUtils.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Aggregates numeric values using either sum or average based on specified strategy
+ * @param {Array<number>} values - Array of numbers to aggregate
+ * @param {'sum' | 'mean'} strategy - Aggregation strategy, either 'sum' or 'mean';
+ * add a custom strategy by modifying the aggregators object in aggregationFunctions.js
+ * and updating JSDoc literal comment
+ * @returns {number} The aggregated value
+ */
+export function aggregateValues(values, strategy = "sum") {
```
nikhilwoodruff, line 28: OK I'm going to walk back my lack of preference and say default should be `last`, and definitely not `sum`.

Poverty impacts, inequality impacts, winners and losers, pretty much all impacts will generate nonsensical results when you do `sum`.

nikhilwoodruff (approved):
These tests look sensible. Just one change- after that, good to merge.

src/api/societyWideAggregation/aggregateUtils.js
MaxGhenis: Isn't this only for budgetary impacts now, where we need to sum?

MaxGhenis: what about renaming the function to sumValues and removing the option to do anything different?

nikhilwoodruff: Actually yes I don't think we need to change this because I just checked out anthony's main UI PR and all other impacts aren't clickable.

MaxGhenis: I still would suggest removing the mean strategy. We should never take the mean across years and so we should not provide the functionality to do so. If one of you agree please file an issue.

anth-volk: Makes sense. Opened as https://github.com/PolicyEngine/policyengine-app/issues/2553.

PR #2503, opened 2025-04-20
Fixed time zone issue caused by formatFullDate()

### LINKED ISSUE #2447: Investigate whether `formatFullDate` causes time zone issues
Similar to #2417. We need to check whether the `formatFullDate` function formats a date and includes a time zone, thereby causing formatted dates to sometimes show up as one day earlier than intended for users in time zones west of UTC. This is the snippet from Mike as part of review for #2430:

Verified these are the only two cases of getFullYear in the app, but I also saw this: https://github.com/PolicyEngine/policyengine-app/blob/d46cbcabaf2f6999807e9b7aa02b3dd9d8747f7e/src/lang/format.js#L85-L90

Do we need to change that too?

v-lerie:
## Description

Fixes #2447.

## Changes

- Modified formatFullDate() to parse the date as UTC and use the timeZone: 'UTC' option in Intl.DateTimeFormat().
- This change ensures that the date is formatted correctly for all users, regardless of their local time zone.

## Tests

- Wrote a test using formatFullDate("2023-01-01", "US") and confirmed it returned "December 31, 2022" due to local time zone offset.

### DIFF
--- a/src/lang/format.js
+++ b/src/lang/format.js
@@ -85,6 +85,7 @@ export function formatPercent(number, countryId, options) {
 export function formatFullDate(date, countryId, options) {
   return new Intl.DateTimeFormat(localeCode(countryId), {
     dateStyle: "long",
+    timeZone: "UTC",
     ...options,
   }).format(new Date(date));
 }


anth-volk (approved):
I don't believe this contains any tests, despite the description indicating it, but this will definitely be helpful to have. Thanks!

Issue #2534, opened 2025-05-15
Make it a code section rather than a image


SEEDART007:
this looks so bad in image

![Image](https://github.com/user-attachments/assets/fa8e5566-6f7e-4231-8c6c-905d329185fc)

should be a code block like this which seems great and really usable

![Image](https://github.com/user-attachments/assets/486f74eb-60ab-4350-86f3-8fc1a2a29751)

MaxGhenis:
Thank you for filing this issue. Closing as part of the `policyengine-app` (v1) wind-down: policyengine.org is now served by [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2), and this repository is being archived, so this issue is frozen against a codebase that no longer deploys anywhere.

If this is still relevant on the current site, please re-file it against [`policyengine-app-v2`](https://github.com/PolicyEngine/policyengine-app-v2) — we'd welcome it there. Apologies that this sat open through the migration.

PR #2538, opened 2025-05-15
Image removed and code added for good look

sumitdethe27:
Issue: #2534

## Changes
Removed this Image block and replaced with a codeblock
image

Replace image with code block in API showcase

- Modified `src/pages/home/HomeTransparency.jsx`: removed static image import and added <CodeBlock> with JSON response
- Updated `src/components/ShowcaseItem.tsx` to handle inline code block as child instead of image

## After new changes applied
image

### DIFF
--- a/src/components/CodeBlock.jsx
+++ b/src/components/CodeBlock.jsx
@@ -1,0 +1,20 @@
+import React from "react";
+import CodeEditor from "@uiw/react-textarea-code-editor";
+
+export default function CodeBlock({ code, language = "json" }) {
+  return (
+    <CodeEditor
+      value={code}
+      language={language}
+      padding={15}
+      style={{
+        fontFamily: 'ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace',
+        fontSize: 14,
+        backgroundColor: "#f5f5f5",
+        borderRadius: 8,
+        marginTop: 12,
+      }}
+      readOnly
+    />
+  );
+}
--- a/src/layout/ShowcaseItem.jsx
+++ b/src/layout/ShowcaseItem.jsx
@@ -46,6 +46,7 @@ export default function ShowcaseItem({
         <p>{description}</p>
         <EmphasisedLink url={link} text={linkTitle} />
       </div>
+
       {imageIsMissing ? (
         <div
           style={{
@@ -66,7 +67,7 @@ export default function ShowcaseItem({
             }}
           />
         </div>
-      ) : (
+      ) : typeof image === "string" ? (
         <img
           src={image}
           width={displayCategory === "desktop" ? 400 : "100%"}
@@ -76,7 +77,17 @@ export default function ShowcaseItem({
           }}
           alt={altText}
         />
+      ) : (
+        <div
+          style={{
+            width: displayCategory === "desktop" ? 400 : "100%",
+            marginTop: displayCategory === "desktop" ? 0 : 20,
+          }}
+        >
+          {image}
+        </div>
       )}
+
     </div>
   );
 }
--- a/src/pages/home/HomeTransparency.jsx
+++ b/src/pages/home/HomeTransparency.jsx
@@ -1,6 +1,6 @@
 import style from "../../style";
 import Section from "../../layout/Section";
-import apiScreenshot from "../../images/home/api_screenshot.png";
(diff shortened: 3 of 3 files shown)


src/components/CodeBlock.jsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
```
anth-volk, line 1: __nit:__ This line not required post React 16.8

src/components/CodeBlock.jsx
```diff
@@ -0,0 +1,20 @@
...
+import CodeEditor from "@uiw/react-textarea-code-editor";
```
anth-volk, line 2: __question, blocking:__ I thought we already had a code block component? If so, I'd recommend using that, or, if it can't be modified to fit this need, communicating why in the name of the component (e.g., `CodeBlockSmall`, `CodeBlockImage`, etc.)

src/components/CodeBlock.jsx
```diff
@@ -0,0 +1,20 @@
...
+      value={code}
+      language={language}
+      padding={15}
+      style={{
+        fontFamily: 'ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace',
+        fontSize: 14,
+        backgroundColor: "#f5f5f5",
+        borderRadius: 8,
+        marginTop: 12,
+      }}
```
anth-volk, line 16: __issue:__ I'd change this styling to better reflect our design

Check out https://policyengine.org/us/api for an example of a code block we already have. In `styles` we have a series of color constants; please use those here.

src/layout/ShowcaseItem.jsx
```diff
@@ -66,7 +67,7 @@ export default function ShowcaseItem({
             }}
           />
         </div>
-      ) : (
+      ) : typeof image === "string" ? (
```
anth-volk, line 70: __issue, blocking:__ Usage here is unclear

I think what you're shooting for here is to treat the code editor as an `image` prop, albeit it's not an image address, so you've added a third option. However, that's unclear from the code. At this point, if we're doing a three-option "image" input, I'd recommend creating a new component and adding a "type" prop to determine how to display.

src/layout/ShowcaseItem.jsx
```diff
@@ -76,7 +77,17 @@ export default function ShowcaseItem({
...
           alt={altText}
         />
+      ) : (
+        <div
+          style={{
+            width: displayCategory === "desktop" ? 400 : "100%",
+            marginTop: displayCategory === "desktop" ? 0 : 20,
+          }}
+        >
+          {image}
```
anth-volk, line 87: __issue:__ We need to determine how to handle alt text for accessibility purposes, since this isn't an image but we're treating it as one

src/pages/home/HomeTransparency.jsx
```diff
@@ -19,24 +19,34 @@ export default function HomeTransparency() {
...
       />
-      <ShowcaseItem
-        title="Our models are tested extensively"
-        description="PolicyEngine's software undergoes thousands of automated tests every day to ensure our simulations produce accurate results."
-        linkTitle="Explore the documentation"
-        link={`https://policyengine.github.io/policyengine-${countryId}`}
-        image={testingScreenshot}
-        altText="Screenshot of PolicyEngine's test coverage"
-        borderColor={style.colors.BLACK}
-      />
```
anth-volk, line 30: __issue, blocking:__ Please do not remove this block. Testing is at the core of what we do.

src/pages/home/HomeTransparency.jsx
```diff
@@ -19,24 +19,34 @@ export default function HomeTransparency() {
...
+          "income_tax_pre_charges": {
+            "2023": 3486
+          },
+          "is_CTC_child_limit_exempt": {
+            "2023": true
+          },
+          "is_QYP": {
+            "2023": false
+          }
+        }`}
```
anth-volk, line 42: __minor:__ For code quality/maintenance purposes, please define this as a separate JS object, then pass by variable

Unfortunately, these params are also out-of-date, but I can give you modern ones in another review round

anth-volk (changes requested):
Thanks for your changes @sumitdethe27. I've highlighted some issues, including some blocking.

anth-volk:
Closing due to staleness

PR #2546, opened 2025-05-19
i button is now clickable

SEEDART007:
Fixes #2528

https://github.com/user-attachments/assets/ceb000b0-2b43-46a2-b356-fade7921945a


SEEDART007:
@anth-volk done

PR #2551, opened 2025-05-19
Redirect enhanced_us region requests and remove deprecated `enhanced_us` code

### LINKED ISSUE #2549: Redirect `region=enhanced_us` URL param to `region=us&dataset=enhanced_cps`
We should take any inbound URL that contains the deprecated `region=enhanced_us` option and change the search params to `region=us&dataset=enhanced_cps` in order to finally move away from this deprecated param. We should also remove all code handling `reigon=enhanced_us`, EXCEPT the code filtering `enhanced_us` out of the API's regions list. In two future PRs, we should remove that region option from the API, then remove the filter, to prevent users accidentally enabling it.

anth-volk:
## Description

Fixes #2549

## Changes

* Redirects all inbound requests containing the deprecated URL parameter `region=enhanced_us` to `region=us&dataset=enhanced_cps`
* Removes all code referencing the `region=enhanced_us` setting, except the code filtering the `enhanced_us` region out of the API's regions list; this will be addressed after removing the option from the API

## Screenshots

N/A

## Tests

Added tests for new function to update deprecated URL parameters and to ensure that `PolicyEngine.jsx` properly modifies said deprecated parameters when rendering.

### DIFF
--- a/src/__tests__/PolicyEngine.test.js
+++ b/src/__tests__/PolicyEngine.test.js
@@ -1,0 +1,84 @@
+import { BrowserRouter, useSearchParams } from "react-router-dom";
+import { render } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import PolicyEngine from "../PolicyEngine";
+
+jest.mock("react-router-dom", () => {
+  const originalModule = jest.requireActual("react-router-dom");
+  return {
+    __esModule: true,
+    ...originalModule,
+    useSearchParams: jest.fn(),
+  };
+});
+
+global.fetch = jest.fn(() =>
+  Promise.resolve({
+    json: () => Promise.resolve({}),
+    ok: true,
+  }),
+);
+
+afterEach(() => {
+  global.fetch.mockClear();
+});
+
+describe("PolicyEngine.jsx", () => {
+  test("Renders for US if proper data passed", () => {
+    useSearchParams.mockImplementation(() => [
+      new URLSearchParams(),
+      jest.fn(),
+    ]);
+
+    const { getByText } = render(
+      <BrowserRouter>
+        <PolicyEngine />
+      </BrowserRouter>,
+    );
+
+    expect(
+      getByText("Computing Public Policy for Everyone"),
+    ).toBeInTheDocument();
+    expect(getByText("Trusted across the US")).toBeInTheDocument();
+  });
+  test("Converts deprecated 'region=enhanced_us' URL search param to 'region=us' and 'dataset=enhanced_cps'", () => {
+    const deprecatedEnhancedUsParams = new URLSearchParams({
+      region: "enhanced_us",
+      country_id: "us",
+    });
+
+    const updatedSearchParams = new URLSearchParams({
+      country_id: "us",
+      region: "us",
+      dataset: "enhanced_cps",
+    });
+
+    const mockSearchParams = {
+      get: jest.fn((key) => deprecatedEnhancedUsParams.get(key)),
+      has: jest.fn((key) => deprecatedEnhancedUsParams.has(key)),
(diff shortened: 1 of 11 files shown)


src/data/reformDefinitionCode.js
```diff
@@ -177,22 +177,15 @@ export function getImplementationCode(
   // Check if the region has a dataset specified
   const hasDatasetSpecified = Object.keys(DEFAULT_DATASETS).includes(dataset);
 
-  const COUNTRY_LEVEL_US_REGIONS = ["us", "enhanced_us"];
-
-  const isState =
-    countryId === "us" && !COUNTRY_LEVEL_US_REGIONS.includes(region);
+  const isState = countryId === "us" && region !== "us";
```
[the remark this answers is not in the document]
anth-volk: We are purposely trying to fully deprecate `enhanced_us`, hence the removal

src/__tests__/PolicyEngine.test.js
```diff
@@ -0,0 +1,84 @@
...
+    ).toBeInTheDocument();
+    expect(getByText("Trusted across the US")).toBeInTheDocument();
+  });
+  test("Converts deprecated 'region=enhanced_us' URL search param to 'region=us' and 'dataset=enhanced_cps'", () => {
+    const deprecatedEnhancedUsParams = new URLSearchParams({
+      region: "enhanced_us",
+      country_id: "us",
+    });
+
+    const updatedSearchParams = new URLSearchParams({
```
mikesmit, line 50: **Comment, non-blocking** This seems like a lot of work. Is there no way to just check the URL has the right search params instead of having to mock the setSearchParams stuff?

src/routing/updateDeprecatedSearchParams.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+export function updateDeprecatedSearchParams(searchParams) {
+  // Check if any deprecated params are present
+  const hasDeprecatedParams = DEPRECATED_SEARCH_PARAMS.some(
+    (param) =>
+      searchParams.has(param.oldKey) &&
+      searchParams.get(param.oldKey) === param.oldValue,
+  );
+  if (!hasDeprecatedParams) {
+    return null;
```
mikesmit, line 43: **NOn-blocking, suggestion/question** I'm like... 78% sure that setSearchParams is smart enough to know if nothing changed and not reload the page. Is that not true?

If so, you can just return the searchParams unchanged.

src/routing/updateDeprecatedSearchParams.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  let newSearchParams = copySearchParams(searchParams);
```
mikesmit, line 46: **Accolade** Thank you for copying before editing.

mikesmit (approved):
A few comments. None blocking.

src/routing/updateDeprecatedSearchParams.js
anth-volk: The way I wrote this, the main container component `PolicyEngine.jsx` invokes this function every time it's rendered, then takes the output and if it's not null, sets the resulting object as the new search params. I'm not sure if it's because of that explicit invocation, but in testing, `setSearchParams` always reset the params, even when duplicate.

src/__tests__/PolicyEngine.test.js
anth-volk: Unfortunately, looks like since React Router v6, any solution that involves explicitly confirming the URL requires something like this.

src/data/reformDefinitionCode.js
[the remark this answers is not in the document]
nikhilwoodruff: lol

PR #2555, opened 2025-05-20
Remove enhanced_us region filter

### LINKED ISSUE #2550: Remove code filtering the `enhanced_us` option out of the regions list for US policy simulations
Needs to be done following merging and deployment of https://github.com/PolicyEngine/policyengine-api/issues/2486. After the deprecated `enhanced_us` region option is removed from the API, we can remove the unnecessary code to filter it out from the regions we display to users using the US site. This should be a very small removal of a couple of lines.

anth-volk:
## Description

Fixes #2550.

## Changes

Removes the code filtering `enhanced_us` out of the region options for society-wide simulations, as the region no longer exists in the API.

## Screenshots

N/A

## Tests

N/A

### DIFF
--- a/src/pages/policy/PolicyRightSidebar.jsx
+++ b/src/pages/policy/PolicyRightSidebar.jsx
@@ -34,9 +34,6 @@ function RegionSelector(props) {
     return { value: region.name, label: region.label };
   });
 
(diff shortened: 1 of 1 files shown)


mikesmit (approved):
LGTM.

PR #2558, opened 2025-05-21
countryId() function added in About.jsx

### LINKED ISSUE #1486: About page should use `useCountryId` instead of explicitly popping URL param at index 1

SEEDART007:
Fixes #1486

### DIFF
--- a/src/layout/Header.jsx
+++ b/src/layout/Header.jsx
@@ -599,7 +599,7 @@ const MobileStandardLink = ({ link, countryId, isOpen, index }) => {
       <HoverBox
         hoverStart="left"
         size="100vw"
-        hoverBackgroundColor={style.colors.WHITE}
+        hoverBackgroundColor={style.colors.BLUE_PRESSED}
         style={{
           margin: 30,
         }}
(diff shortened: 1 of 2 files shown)


src/layout/Header.jsx
```diff
@@ -599,7 +599,7 @@ const MobileStandardLink = ({ link, countryId, isOpen, index }) => {
       <HoverBox
         hoverStart="left"
         size="100vw"
-        hoverBackgroundColor={style.colors.WHITE}
+        hoverBackgroundColor={style.colors.BLUE_PRESSED}
```
anth-volk, line 602: __issue, blocking:__ The changes to this file originate from a different feature branch; please remove and only keep the changes relevant to the feature

anth-volk (changes requested):
@SEEDART007 Please remove the fragments from a different feature branch and rebase off `master`.

anth-volk:
Closing due to staleness

PR #2562, opened 2025-05-23
Fix: solve duplicate key error in simulation

### LINKED ISSUE #2554: React duplicate key error in simulation queue page
E.g.:
```
Warning: Encountered two children with the same key, `an hour ago`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version. Component Stack: 
```

This occurs when items in a React Array are given a `key` prop that is not explicitly unique

miguelro20:
Fixes #2554 

## Description

The fix was made to prevent an error that happened in simulation page because of duplicate keys.

## Changes

Changed the key used inside the table from start-time to unique key. The unique key was already included in the code, the fix was to add it in the simulation object.

### DIFF
--- a/src/pages/Simulations.jsx
+++ b/src/pages/Simulations.jsx
@@ -49,7 +49,12 @@ export default function SimulationsPage() {
           const uniqueSimulations = new Map();
 
           data.result.forEach((simulation) => {
+            // Creates unique key for each simulation object
             const key = `${simulation.reform_policy_id}-${simulation.baseline_policy_id}-${simulation.region}-${simulation.time_period}-${simulation.options_json}`;
+            
+            // Store it on the simulation object 
+            simulation.rowKey = key;
+            
             if (!uniqueSimulations.has(key)) {
               uniqueSimulations.set(key, simulation);
             } else {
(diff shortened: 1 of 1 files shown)


src/pages/Simulations.jsx
```diff
@@ -119,7 +124,7 @@ export default function SimulationsPage() {
         </p>
       </PageHeader>
       <Section>
-        <Table dataSource={data} columns={columns} rowKey="start_time" />
+        <Table dataSource={data} columns={columns} rowKey="rowKey" />
```
anth-volk, line 127: __issue, blocking:__ Per Ant Design's docs, the `rowKey` prop defaults to a given row's `key` value; is the fix not as simple as removing the original `rowKey="start_time"` designation?

anth-volk (changes requested):
Thanks for your changes @miguelro20. Had a question for you before proceeding.

anth-volk (approved):
Your explanation sounds good to me. LGTM pending test passage.

anth-volk:
Closing due to staleness

PR #2592, opened 2025-05-30
Fix app white screens when used on the phone

### LINKED ISSUE #2591: App white screens when used on the phone
Example when reducing the window size:

https://github.com/user-attachments/assets/16d50e2e-58b0-41ea-9e9f-c625a66da4dc

MaxGhenis:
Fixes #2591

### DIFF
--- a/src/layout/Responsive.jsx
+++ b/src/layout/Responsive.jsx
@@ -17,10 +17,17 @@ const determineDisplayCategory = (width) => {
 
 const useDisplayCategory = () => {
   const [currentDisplayCategory, setcurrentDisplayCategory] = useState(
-    determineDisplayCategory(window.innerWidth),
+    typeof window !== "undefined"
+      ? determineDisplayCategory(window.innerWidth)
+      : "desktop", // Default to desktop during SSR
   );
 
   useEffect(() => {
+    // Set the correct display category on mount
+    if (typeof window !== "undefined") {
+      setcurrentDisplayCategory(determineDisplayCategory(window.innerWidth));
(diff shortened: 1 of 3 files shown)


src/layout/Responsive.jsx
```diff
@@ -17,10 +17,17 @@ const determineDisplayCategory = (width) => {
 
 const useDisplayCategory = () => {
   const [currentDisplayCategory, setcurrentDisplayCategory] = useState(
-    determineDisplayCategory(window.innerWidth),
+    typeof window !== "undefined"
+      ? determineDisplayCategory(window.innerWidth)
+      : "desktop", // Default to desktop during SSR
```
anth-volk, line 22: __nit, non-blocking:__ We don't use SSR, so not sure how this check is relevant

src/layout/Responsive.jsx
```diff
@@ -32,10 +39,19 @@ const useDisplayCategory = () => {
 
 const useMobile = () => {
   const [currentDisplayCategory, setcurrentDisplayCategory] = useState(
-    determineDisplayCategory(window.innerWidth) === "mobile",
+    typeof window !== "undefined"
+      ? determineDisplayCategory(window.innerWidth) === "mobile"
+      : false, // Default to false (non-mobile) during SSR
```
anth-volk, line 44: __nit, non-blocking:__ Here again, we don't use server-side rendering

anth-volk (approved):
Left a couple nits in here, but otherwise, this all works when I test locally.

PR #2596, opened 2025-06-02
Added abolitions filter in API Variable/Parameter Explorer

harnish7576:
Fixes Issue 2488 

## Description
Adds toggle-based filtering to show or hide parameters related to "abolition" in the parameter explorer.

## Changes
- Introduced filterByAbolition toggle logic.
- Filter excludes gov.abolitions* items unless toggle is enabled.

## Screenshots
Screenshot 2025-05-29 at 2 02 34 PM

### DIFF
--- a/src/pages/learn/APIGeneralContent.jsx
+++ b/src/pages/learn/APIGeneralContent.jsx
@@ -399,6 +399,8 @@ export function VariableParameterExplorer(props) {
   const { metadata } = props;
   const [query, setQuery] = useState("");
   const [selectedCardData, setSelectedCardData] = useState(null);
+  const [filterByAbolition, setFilterByAbolition] = useState(false);
+
   const [page, setPage] = useState(0);
 
   const MAX_ROWS = 3;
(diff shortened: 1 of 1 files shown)


src/pages/learn/APIGeneralContent.jsx
```diff
@@ -399,6 +399,8 @@ export function VariableParameterExplorer(props) {
   const { metadata } = props;
   const [query, setQuery] = useState("");
   const [selectedCardData, setSelectedCardData] = useState(null);
+  const [filterByAbolition, setFilterByAbolition] = useState(false);
```
anth-volk, line 402: __issue, minor, blocking:__ I'd prefer a minor re-frame of this feature

I think I'd prefer to see this as an opt-in; The checkbox would read "Show abolition parameters," ideally with a hoverable question mark to the right that contains a tooltip describing what an abolition is. I'd probably use the following language preliminarily:

"Abolition parameters are used to remove all impacts from a standard parameter, usually by setting its value to 0"

I'd have this default to unchecked, then when a user checks it, the abolition parameters are displayed.

anth-volk (changes requested):
Thanks @harnish7576, this is a great start. I wanted to propose a subtle reframing of the feature, then review visuals next time around.

MaxGhenis:
Please use the keyword as needed to link https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

src/pages/learn/APIGeneralContent.jsx
harnish7576: Done! I've made the requested changes and updated the screenshot above to show the new behavior.

src/pages/learn/APIGeneralContent.jsx
```diff
@@ -408,26 +411,39 @@ export function VariableParameterExplorer(props) {
...
+      );
+    }
+
+    return true;
   };
 
   const parameterCards = Object.values(metadata.parameters || {})
     .filter((p) => p.type === "parameter")
-    .filter((p) => p.label && !/^\d+$/.test(p.label)) // Add this filter
+    .filter((p) => p.label && !/^\d+$/.test(p.label))
```
anth-volk, line 441: __issue, blocking:__ I think we need to remove the filter that doesn't display numeric-only params

I believe you added this because I made a comment that it'd be great to not just end up with a bunch of params called "1" at the beginning of every search result. However, until we determine a good way of actually displaying these things, I think we need to display them.

src/pages/learn/APIGeneralContent.jsx
```diff
@@ -408,26 +411,39 @@ export function VariableParameterExplorer(props) {
...
     .filter(filterByQuery)
     .map((p) => ({ ...p, type: "parameter" }));
 
   const variableCards = Object.values(metadata.variables || {})
-    .filter((v) => v.label && !/^\d+$/.test(v.label)) // Add this filter
+    .filter((v) => v.label && !/^\d+$/.test(v.label))
```
anth-volk, line 446: __issue, blocking:__ Here, too, we should remove the digit filter

anth-volk (changes requested):
@harnish7576 Thanks for your continued work on this. There's one minor blocking change, then we'll be good to merge.

src/pages/learn/APIGeneralContent.jsx
harnish7576: Changed it back to as it was. It now looks like
Screenshot 2025-06-23 at 2 09 18 PM

Issue #2598, opened 2025-06-02
change date on manager's amendment post to may 24


MaxGhenis:
bill came out may 22, we posted may 24

PR #2618, opened 2025-06-23
Fixed user profile page error

### LINKED ISSUE #2589: The profile page seems to break periodically
https://github.com/user-attachments/assets/afb848ec-5796-4f15-8249-bc1dd0a4d31f

The screen recording is from May 29th but it worked when accessing on May 30th, not sure if this was fixed because I logged out and logged back in a day later

v-lerie:
## Description

Fixes #2589 

## Changes

- Updated the logic for loading the user profile page to prevent errors when a user profile does not exist yet.
- This fix is paired with the backend fix in [policyengine-api PR #2598](https://github.com/PolicyEngine/policyengine-api/pull/2598), which adds logic to automatically create a user profile during a GET /user-profile call if none exists.

### DIFF
--- a/src/api/call.js
+++ b/src/api/call.js
@@ -4,7 +4,8 @@ import { buildVariableTree, getTreeLeavesInOrder } from "./variables";
 import { wrappedJsonStringify, wrappedResponseJson } from "../data/wrappedJson";
 import { useAuthenticatedFetch } from "../hooks/useAuthenticatedFetch";
 
-const POLICYENGINE_API = "https://api.policyengine.org";
+const POLICYENGINE_API =
+  process.env.REACT_APP_API_URL || "https://api.policyengine.org";
 
 /**
  * returns an api call function that can be used to make requests
--- a/src/pages/UserProfilePage.jsx
+++ b/src/pages/UserProfilePage.jsx
@@ -318,7 +318,9 @@ function UserProfileSection(props) {
   } else if (dispState === STATES.EMPTY) {
     dispUserSince = "Loading";
   } else {
-    dispUserSince = dateFormatter.format(accessedUserProfile.user_since);
+    dispUserSince = dateFormatter.format(
+      new Date(parseInt(accessedUserProfile.user_since) * 1000),
+    );
   }
 
   let dispCountry = "";


src/pages/UserProfilePage.jsx
```diff
@@ -318,7 +318,9 @@ function UserProfileSection(props) {
   } else if (dispState === STATES.EMPTY) {
     dispUserSince = "Loading";
   } else {
-    dispUserSince = dateFormatter.format(accessedUserProfile.user_since);
+    dispUserSince = dateFormatter.format(
+      new Date(parseInt(accessedUserProfile.user_since) * 1000),
```
anth-volk, line 322: __question, blocking:__ Why multiply by 1000?

src/api/call.js
```diff
@@ -4,7 +4,8 @@ import { buildVariableTree, getTreeLeavesInOrder } from "./variables";
 import { wrappedJsonStringify, wrappedResponseJson } from "../data/wrappedJson";
 import { useAuthenticatedFetch } from "../hooks/useAuthenticatedFetch";
 
-const POLICYENGINE_API = "https://api.policyengine.org";
+const POLICYENGINE_API =
+  process.env.REACT_APP_API_URL || "https://api.policyengine.org";
```
anth-volk, line 8: __question:__ Why add this environment variable?

anth-volk (commented):
A couple questions for you @v-lerie. Thanks for this!

src/pages/UserProfilePage.jsx
v-lerie: Without multiplying by 1000, the date kept showing up as January 1st, 1970. This happened because `accessedUserProfile.user_since` is a Unix timestamp in seconds, but JavaScript’s `Date` expects timestamps in milliseconds. So we multiply by 1000 to format the date correctly.

src/api/call.js
v-lerie: I added `REACT_APP_API_URL` for local development and testing. I wasn’t able to get the frontend and backend to connect properly without it. It defaults to the live API if not set, so production behavior isn’t affected. I left it in since it might be helpful for others running things locally, but I can remove it if it’s not needed.

src/pages/UserProfilePage.jsx
anth-volk: __question, blocking:__ But when using this code with working profiles, I'm able to properly display the date without multiplying by 1000

When we actually write the record to the table, I believe we use JS `Date`, not something native to the database, for this exact purpose.

MaxGhenis:

PR #2629, opened 2025-06-29
Remove mean strategy from aggregateValues function

### LINKED ISSUE #2553: Remove "mean" strategy from `aggregateValues` function
Described in a comment in https://github.com/PolicyEngine/policyengine-app/pull/2499. We should remove the `mean` strategy as an aggregation strategy within `aggregateValues`, leaving for the time being only the `sum` strategy.

SongmingFan123:
Fixes #2553

### DIFF
--- a/src/__tests__/api/societyWideAggregation/aggregateUtils.test.js
+++ b/src/__tests__/api/societyWideAggregation/aggregateUtils.test.js
@@ -5,12 +5,10 @@ import {
 
 import {
   emptyInput,
-  validBaselineReformData,
-  expectedBaselineReformSumResult,
-  expectedBaselineReformMeanResult,
-  validAggregateValuesData,
   expectedAggregateValuesSumResult,
-  expectedAggregateValuesMeanResult,
+  expectedBaselineReformSumResult,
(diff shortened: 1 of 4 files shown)


src/__tests__/api/societyWideAggregation/aggregateUtils.test.js
```diff
@@ -5,12 +5,10 @@ import {
 
 import {
   emptyInput,
-  validBaselineReformData,
+  expectedAggregateValuesSumResult,
```
anth-volk, line 8: __comment:__ Here again, I don't understand why you moved imports around

Technically, it is an improvement, but in doing so, reviewing becomes more challenging because you increase the number of lines changed. Please only make changes to lines of code within the scope of the issue, unless it's necessary/there's a compelling reason.

anth-volk (approved):
Thanks for this @SongmingFan123, this looks good. One comment: please only change lines that are within the scope of the issue unless there is a compelling reason. There is one file that just rearranges certain imports, increasing review time for a change that's technically slightly better, but not compelling enough to include as part of this PR.

PR #2672, opened 2025-07-14
Set benefit increases to negative values in net revenue impact table

### LINKED ISSUE #2601: Set benefit increases to negative values in net revenue impact 5y table
Benefit increases currently show as positive values in the 5y chart ([example](https://policyengine.org/uk/policy?reform=74146&focus=policyOutput.budgetaryImpact&region=uk&timePeriod=2025&baseline=1&uk_local_areas_beta=true&simYears=5)
). But the chart title indicates it's net revenue impact, not net <value of row>. We should set it to negative to correspond to the chart title and have the total row sum the component rows.
Image

v-lerie:
## Description

Fixes #2601 

## Screenshots

image

### DIFF
--- a/src/pages/policy/output/budget/MultiYearBudgetaryImpact.jsx
+++ b/src/pages/policy/output/budget/MultiYearBudgetaryImpact.jsx
@@ -114,10 +114,13 @@ export default function MultiYearBudgetaryImpact(props) {
         item.budgetKey ? item.budgetKey : null,
         item.formula ? item.formula : null,
       ),
+
       yearRange: roundToBillions(
-        item.budgetKey
-          ? impact.budget[item.budgetKey]
-          : item.formula(impact.budget),
+        item.budgetKey === "benefit_spending_impact"
+          ? -Math.abs(impact.budget[item.budgetKey])
+          : item.budgetKey
+            ? impact.budget[item.budgetKey]
+            : item.formula(impact.budget),
         roundingPrecisionByCountry[metadata.countryId] ||
           roundingPrecisionByCountry.default,
       ),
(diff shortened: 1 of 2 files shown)


src/pages/policy/output/budget/MultiYearBudgetaryImpact.jsx
```diff
@@ -114,10 +114,13 @@ export default function MultiYearBudgetaryImpact(props) {
         item.budgetKey ? item.budgetKey : null,
         item.formula ? item.formula : null,
       ),
+
       yearRange: roundToBillions(
-        item.budgetKey
-          ? impact.budget[item.budgetKey]
-          : item.formula(impact.budget),
+        item.budgetKey === "benefit_spending_impact"
+          ? -Math.abs(impact.budget[item.budgetKey])
```
anth-volk, line 120: __issue, blocking:__ I believe this only half-solves the problem

In cases where the government budget loses revenue (e.g., a new benefit program is introduced), the benefits row should be negative, but in cases where the government budget gains revenue (e.g., benefit cuts), it should be positive. This code will always make it positive.

anth-volk, line 120: __issue, blocking:__ It's tough to read this nested conditional

Addressing my other comment, can you modify this to be more readable/CLEAN?

src/pages/policy/output/budget/MultiYearBudgetaryImpact.jsx
```diff
@@ -198,6 +201,10 @@ export function getYearlyImpacts(
       impact = item.result.budget[budgetKey];
     }
 
+    if (budgetKey === "benefit_spending_impact") {
+      impact = -Math.abs(impact);
```
anth-volk, line 205: __issue, blocking:__ Here, too budgetary impacts are always negative when they should be negative only sometimes

anth-volk (changes requested):
Thanks for your work on this @v-lerie, just had a few changes for you.

MaxGhenis:

If v2's revenue table still shows this, please re-file — sounds like a real display bug.

PR #2691, opened 2025-07-23
Hardcode OBBBA iframe URL to fix persistent deep link issues

MaxGhenis:
## Summary

This PR takes a direct approach to fixing the persistent deep link issues with the OBBBA household explorer by hardcoding the correct iframe URL.

## Background

Despite multiple attempts to fix deep links:
- PR #2688 added configuration infrastructure
- PR #2690 fixed a duplicate path bug

Deep links like https://policyengine.org/us/obbba-household-explorer?household=46067&baseline=tcja-expiration still don't work in production.

## Root Cause Analysis

After extensive debugging, it appears that either:
1. An environment variable is being set during build/deployment that overrides our configuration
2. There's a caching issue preventing the new code from being served
3. Some other build-time configuration is interfering

## The Solution

This PR takes the most direct approach:
- **Hardcodes** the correct URL directly in the component
- **Removes** the environment configuration that wasn't working reliably
- **Adds debug logging** to help diagnose any remaining issues

## Changes

1. Hardcoded `baseUrl` to `"https://policyengine.github.io/obbba-scatter"`
2. Removed the `src/config/environment.js` file
3. Added console logging to show the constructed iframe URL

## Testing

After deployment, visiting https://policyengine.org/us/obbba-household-explorer?household=46067&baseline=tcja-expiration should:
1. Log the iframe URL construction in the browser console
2. Pass parameters correctly to the iframe
3. Select and highlight the specified household

## Note

This is admittedly a brute-force solution, but given that the more elegant configuration-based approaches haven't worked in production, this ensures the feature will work for users.


PavelMakarchuk:
Superseded by #2692

PR #2704, opened 2025-07-24
Use conditional chaining to prevent MarkdownFormatter breakage

### LINKED ISSUE #2664: Household AI feature crashes unexpectedly
When using the household AI feature, the app crashes completely. This appears to be driven by something within the `MarkdownFormatter`, as when the error occurs, I get the following error message in the console:

```
TypeError: can't access property "split", t is undefined
    h2 MarkdownFormatter.jsx:478
    React 7
    _ scheduler.production.min.js:13
    E scheduler.production.min.js:14
<anonymous code>:1:145535
```

anth-volk:
## Description

Fixes #2664.

## Changes

In a couple spots, the `MarkdownFormatter` attempts to access attributes or methods that may not be present if a given input is `undefined`. This includes the header formatting function, which attempts to take a string and use `String.split()` to divide the header into segments before styling, but if a given input is not a string, `MarkdownFormatter` crashes.

This PR uses optional chaining to prevent this error in a few locations that use `String.split()`. It leaves out the line reading:
```
    [leftContent, rightContent] = content.split("&&&");
```
As this will actually resolve to `undefined` if optional chaining is employed due to the array spread operator.

These changes do not remove `MarkdownFormatter`. Removing `MarkdownFormatter` would first require minor API-side changes, as the prompt we employ instructs Claude to return code in markdown. @MaxGhenis if you'd still prefer that we remove `MarkdownFormatter`, I can do so, but would recommend first pushing this to re-enable the feature.

## Screenshots

https://github.com/user-attachments/assets/ec02098b-7f10-4ac8-ae78-29906cd3b2ac

## Tests

N/A

### DIFF
--- a/src/layout/MarkdownFormatter.jsx
+++ b/src/layout/MarkdownFormatter.jsx
@@ -369,7 +369,7 @@ export function MarkdownFormatter({ markdown, backgroundColor, dict, pSize }) {
           // If href=#user-content-fn-1, id should be user-content-fnref-1 and vice versa
           if (href.startsWith("#user-content-fn-")) {
             id = href.replace("#user-content-fn-", "user-content-fnref-");
-            footnoteNumber = parseInt(id?.split("-").pop());
+            footnoteNumber = parseInt(id?.split("-").pop() || "0");
           } else if (href.startsWith("#user-content-fnref-")) {
             id = href.replace("#user-content-fnref-", "user-content-fn-");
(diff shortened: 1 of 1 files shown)


MaxGhenis (approved):
LGTM and good to see we (hopefully) don't have to drop markdown. But PTAL at Copilot's suggestions.

PR #2719, opened 2025-08-05
Fixed issue 1936: brackets sorting

### LINKED ISSUE #1936: Brackets do not appear in chronological order if more than 10
image

AnushkaNilangekar:
Fixes #1936 

## Description

This PR fixes the incorrect ordering of bracketed parameter nodes (e.g., "Bracket 10" appearing before "Bracket 2") in the parameter tree sidebar. The issue was caused by missing index assignments for bracket nodes in the parameter tree construction, and by sorting logic that did not distinguish between bracket and non-bracket nodes. This change ensures bracket nodes are always sorted numerically, while all other nodes are sorted alphabetically.

## Changes

_Parameter Tree Construction (parameters.js)_

Assigns a numeric index to bracket nodes during tree construction, based on their bracket number.
Non-bracket nodes retain their default or module-based index.

_Tree Sorting Logic (PolicyPage.jsx):_

Updates the sortTreeInPlace function to:
Detect if all nodes at a given tree level are bracket nodes (using a regex on the label).
Sort bracket nodes numerically by their index property.
Sort all other nodes alphabetically by their label.

## Screenshots

https://www.loom.com/share/cec71fa84dae43e6b58b7e08b49782d1?sid=3d6c2f5a-b909-4bdf-9dbd-790bf173b26d

## Tests

Manually verified in the UI that bracketed parameters are now always sorted numerically, even for double-digit brackets. Confirmed that non-bracket nodes remain sorted alphabetically.

### DIFF
--- a/src/api/parameters.js
+++ b/src/api/parameters.js
@@ -31,8 +31,10 @@ export function buildParameterTree(parameters) {
         (cumulativePath in parameters && parameters[cumulativePath].label) ||
         key;
       // Transform e.g. "0]" -> 1
+      let bracketIndex = 0;
       if (key.endsWith("]")) {
-        label = `Bracket ${parseInt(key.slice(0, -1)) + 1}`;
+        bracketIndex = parseInt(key.slice(0, -1));
+        label = `Bracket ${bracketIndex + 1}`;
       }
       label = label.replaceAll("_", " ");
       if (!currentNode.children) {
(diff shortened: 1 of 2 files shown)


anth-volk (approved):
Thanks for these contributions @AnushkaNilangekar! They're a real improvement to the UI. Generally, we'd ask for tests to accompany these, but given that we will be looking to deprecate this module by the end of the year, as well as the narrow scope of the changes, we will skip that this time.

Issue #1607, opened 2024-04-15
Add state income tax post


PavelMakarchuk:
https://blog.policyengine.org/policyengine-launches-state-income-tax-modeling-nationwide-90d15bc2564f

PR #2829, opened 2026-04-17
Use British spelling for UK LSR parameter paths

vahid-ahmadi:
## Summary
- Companion to PolicyEngine/policyengine-uk#1607, which renames `gov.simulation.labor_supply_responses.*` → `gov.simulation.labour_supply_responses.*` in the UK repo.
- Updates the two UK-only path strings in `PolicyRightSidebar.jsx` (`income_elasticity`, `substitution_elasticity`) so the LSR sliders keep resolving after the UK rename lands.
- US paths (`elasticities.*`) are untouched — those live in `policyengine-us` and aren't being renamed.

## Coordination
Merge order matters: this PR should land **immediately before or after** policyengine-uk#1607 to minimize the window where one side disagrees with the other. Safer sequence:
1. Merge this PR first so the app expects the new paths.
2. Merge policyengine-uk#1607, then bump `policyengine-uk` in the app.

## Test plan
- [ ] Run the app against a policyengine-uk build with the rename applied and confirm the LSR toggle still adds both elasticity parameters to the policy.
- [ ] Confirm no regressions to the US LSR flow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -2,3 +2,4 @@
   changes:
     changed:
       - Added household-level impact analysis by pension contributions to UK income tax and NI reforms blog post
+      - Updated UK LSR parameter paths to use British spelling (`labour_supply_responses`) to match policyengine-uk rename.
--- a/src/pages/policy/PolicyRightSidebar.jsx
+++ b/src/pages/policy/PolicyRightSidebar.jsx
@@ -111,10 +111,10 @@ function BehavioralResponseToggle(props) {
   const behavioralResponseReforms = useMemo(
     () => ({
       uk: {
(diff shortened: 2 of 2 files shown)


MaxGhenis:
I reviewed this after merging PolicyEngine/policyengine-uk#1607. I am leaving the app PR unmerged for now because the frontend would start sending the new `labour_supply_responses.*` paths before we know the deployed API/country package has a PE-UK release containing those paths. #1607 keeps the legacy `labor_supply_responses.*` paths alive as aliases, so production is safe without this app change. This can land after the PE-UK release/API update that includes #1607.

MaxGhenis:

v2 may have already normalized these paths — worth a quick check before re-filing.

